PR #4, opened 2016-12-05
design doc

andrayantelo:
Like this? @johntyree

### DIFF
--- a/designdoc.md
+++ b/designdoc.md
@@ -1,0 +1,31 @@
+
+steps usually taken when generating just one month div that
+has the click handler (but not the checkmarks that the user had previously put in
+not doing the save/load thing yet:
+make month object
+initialize it's month state
+generate an empty  month div
+fill the month div
+attach the clickhandler
+loading would occur in place of initializing a month state, since
+loading would give you the month's month state.
+then after attaching the clickhandler, you would generate checkmarks
+
+What should be happening when a user visits this page for the first time?
+it loads, user picks a calendar title, then picks a start date, picks
+how many years he/she wants to track  and clicks on
+a button that says "set". then the page generates a calendar starting at
+the start date provided and going on for the amount of years that the user
+gave. The user can then put a checkmark on each day, making sure to click on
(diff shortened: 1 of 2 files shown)


johntyree:
Yes :)

designdoc.md
```diff
@@ -0,0 +1,31 @@
...
+has the click handler (but not the checkmarks that the user had previously put in
+not doing the save/load thing yet:
+make month object
+initialize it's month state
+generate an empty  month div
+fill the month div
+attach the clickhandler
+loading would occur in place of initializing a month state, since
+loading would give you the month's month state.
+then after attaching the clickhandler, you would generate checkmarks
```
johntyree, line 12: what do you mean "generate checkmarks" here? Via fake clicking using jquery?

johntyree:
So this is good and starts to explain how the site will look and what the user can do, but not how you plan on implementing these things.

The details behind things like "save the calendarState" are where it gets difficult. How does that work?

designdoc.md
andrayantelo: Yeah,  I already have the function that does that. It toggles between classes.

andrayantelo:
I plan on doing it the same way that I did it previously with the first version of this thing. I have a calendarState, which is a dictionary (or an object I guess in JavaScript), and in it you have the startDate (which the user provided), an array of all the years the user wanted to track, an array of monthStates, a listName (a string we get from the input tag where the user types a title for their calendar), the numberOfYears, and the numberOfMonths that will be required for this calendar. So, this is what gets stored, and you can pretty much build the entire calendar using this information. Or at least that's the idea I'm going for. @johntyree not sure if I have to tag you in this so that you know I responded?

andrayantelo:
@johntyree Right now when you check a day, the day number disappears and you just have a checkmark there. Do you think if you wanted to use this calendar you'd want to see the day number still even after you've checked the day?

Did it. Looks great!

johntyree:
This is all out of date now.

PR #60, opened 2017-01-07
got rid of monthStates, added attributes to Month Object

andrayantelo:
Modified month methods to reflect this change. Will probably need to get rid of collectCheckedDays and generateCheckMarks, and attachClickHandler methods from Month object and instead have them be methods on Calendar object . Also each day has an id "YYYYMMDD". 

Am still using self.monthId (the year followed by month index).


calendar_reboot.js
```diff
@@ -17,28 +17,64 @@ $(document).ready(function() {
...
+        
+            //the storage key is the calendar's unique Id
+            var storageKey = calendar.calendarState.uniqueId;
+            
+            //get the calendar title to put in the dropdown
+            var calendarTitle = calendar.calendarState.calendarTitle;
+       
+            //the calendarUniqueId dictionary also needs to be stored in localstorage. use 
+            //global variable masterKey as the storage key
+            storeInLocalStorage(masterKey, calendarUniqueId);
```
johntyree: Why would this have changed? This function doesn't change it.

calendar_reboot.js
```diff
@@ -17,28 +17,64 @@ $(document).ready(function() {
...
+            
+            //update the calendarState (with checkmark information)
+            calendar.updateCalendarAttributes(calendar.calendarState.startDate, calendar.calendarState.numberOfYears, calendar.calendarState.calendarTitle);
+            calendar.updateCalendarState(calendar.monthObjects);
+            
+            //store the uniqueId and calendarTitle inside of an object (dictionary) 
+            //and then save this progress in localStorage
+            calendarUniqueId[calendar.calendarState.uniqueId] = calendarTitle; 
+            storeInLocalStorage(masterKey, calendarUniqueId);
```
johntyree: Oh... we change it down here. Perhaps the first one is a copy-paste error?

calendar_reboot.js
```diff
@@ -354,9 +382,12 @@ $(document).ready(function() {
     }
     
     //load the current_active_calendar
+    //to check, check if active calendar is in localStorage
     var activeCalendarState = loadFromLocalStorage(calendarUniqueId[current_active_calendar]);
     
     if (activeCalendarState !== null) {
+        //create calendar object
+        calendar = new Calendar("", 0, "");
```
johntyree: These empty values, when do they get updated? We set the state below but don't set any other attributes on the calendar object itself?

calendar_reboot.js
```diff
@@ -635,39 +575,42 @@ var Month = function(date) {
...
+            "<h2 id='yearHeader'>" + self.monthYear + "</h2>" +
             "</div>";
         
         //the div ID is the monthID
-        $div.append('<div class="monthframe" id=' + self.monthState.monthId + '></div>');
-        if (self.monthState.monthId == "0" + self.monthState.monthYear.toString()) {
-            $('#' + self.monthState.monthId).append(yearHeader);
+        $div.append('<div class="monthframe" id=' + self.monthId + '></div>');
+       // $div.append('<div class="monthframe"></div>');
+        if (self.monthIndex === "0") {
```
johntyree: Why are the indices strings? Shouldn't they be integers?

andrayantelo: I just changed it. Check it out.

calendar_reboot.js
andrayantelo: I have to refactor the calendar object code, and then I will be able to modify these lines.

calendar_reboot.js
andrayantelo: I kept this one, and deleted the other.

andrayantelo: Yes, I believe it was a copy paste error.

Issue #63, opened 2017-01-08
Rewrite validateInput and validateForm


andrayantelo:

andrayantelo:
Make an error if endDate comes before startDate

andrayantelo:
@johntyree should I have alerts, like if someone leaves title blank, should there be an alert, "Please provide a title" or something like that. An alert from every input, and also an alert if end date is before start date?

johntyree:
Definitely don't spam alerts. I commented in the review, but check out something like http://mdbootstrap.com/javascript/alerts/ for an example of a nice way to tell the user that they need to fix something before they can continue.

johntyree:
Another example http://codeseven.github.io/toastr/demo.html

andrayantelo:
Well,  I am using that thing where the input box turns red. so that should be enough. the only time I have an alert right now is when the end Date comes before the start Date.

johntyree:
When you have a bug like this and a PR that fixes it, you can tie them together by referencing the bug from the PR description or from one of your commit messages.

https://help.github.com/articles/closing-issues-via-commit-messages/

PR #69, opened 2017-01-09
Added function validateForm that checks if the input given for title,…

andrayantelo:
… start date, and end date are not blank, and added validateDates which checks if the end Date comes after the startDate, and I use these functions inside of the SetButton click handler, before the calendar is built.

I also made a function removeFormErrors, so that when you click on clear (which clears the form, it clears the errors too)

Fixes Issue #63

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -89,10 +89,10 @@ $(document).ready(function() {
     
     var validateDates = function(startDateString, endDateString) {
         
-        var startDateMoment = moment(startDateString, "YYYY-MM-DD");
-        var endDateMoment = moment(endDateString, "YYYY-MM-DD");
+        var startDate = moment(startDateString, "YYYY-MM-DD");
+        var endDate = moment(endDateString, "YYYY-MM-DD");
         
-        if (startDateMoment.isBefore(endDateMoment)) {
+        if (startDate.isBefore(endDate)) {
             removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
             return true;
         }
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -144,19 +144,7 @@ $(document).ready(function() {
         
         var isValid = validateTitle && validateStartDate && validateEndDate;
         
-        if (isValid) {
-            var validateEndDate = validateDates(startDateString, endDateString);
-            if (validateEndDate) {
-                return true
-            }
-            else {
-                return false
-            }
-        }
-        else {
-            return false
-        }
-        
+        return (isValid && validateDates(startDateString, endDateString);
        
     };
     
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -87,18 +87,16 @@ $(document).ready(function() {
         removeGlyphicon('#span-titleFormGroup');
     };
     
-    var validateDates = function(startDate, endDate) {
-        console.log("validating the dates");    
+    var validateDates = function(startDateString, endDateString) {
         
-        var startDateMoment = moment(startDate, "YYYY-MM-DD");
-        var endDateMoment = moment(endDate, "YYYY-MM-DD");
+        var startDateMoment = moment(startDateString, "YYYY-MM-DD");
+        var endDateMoment = moment(endDateString, "YYYY-MM-DD");
         
         if (startDateMoment.isBefore(endDateMoment)) {
             removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
             return true;
         }
         else {
-            alert("The end date needs to be after the start date.");
             addFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
             return false;
         }
@@ -106,7 +104,6 @@ $(document).ready(function() {
     };
     
     var validateInput = function(formGroupId) {
-        console.log("validating the input " + formGroupId);
         
         //make a variable for the id string
         var id = formGroupId;
@@ -138,19 +135,29 @@ $(document).ready(function() {
         return true;
     };
     
-    var validateForm = function() {
-        console.log("validating the form");
+    var validateForm = function(startDateString, endDateString) {
+        
         
         var validateTitle = validateInput(calendarTitleId);
         var validateStartDate = validateInput(startDateId);
         var validateEndDate = validateInput(endDateId);
         
-        if (validateTitle && validateStartDate && validateEndDate) {
-            return true;
+        var isValid = validateTitle && validateStartDate && validateEndDate;
+        
+        if (isValid) {
+            var validateEndDate = validateDates(startDateString, endDateString);
+            if (validateEndDate) {
+                return true
+            }
+            else {
+                return false
+            }
         }
         else {
-            return false;
+            return false
         }
+        
+       
     };
     
     
@@ -177,24 +184,23 @@ $(document).ready(function() {
     
     $('#clearButton').click(function() {
         //clear all the entries in the build calendar form
-        console.log("clearing form...");
+        
         removeFormErrors();
         $('#fullForm')[0].reset();
     });
     
     $('#setButton').click(function(){
-        console.log("set Button clicked");
+        
         //take the empty calendar object made when the page is loaded
         // and fill it with the information given in the form
-        //required info are startDate and title. default to tracking 1 
-        //year
+        
         
         var calendarTitle = $("#calendarTitle").val();
         var startDate = $("#startDate").val();
         var endDate = $("#endDate").val();
         
-        if (validateForm() && validateDates(startDate, endDate)) {
-        
+        if (validateForm(startDate, endDate)) {
+            
             //clear the previously displayed calendar <-- or should i reload the page???
             clearPage();
         
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -6,12 +6,11 @@ $(document).ready(function() {
     //will need to use a better key than this for masterKey (if this is
     //even a good way to do this)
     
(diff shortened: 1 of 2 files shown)


calendar_reboot.js
```diff
@@ -66,44 +65,54 @@ $(document).ready(function() {
...
-          //if there is no valid startDate, prompt user for a valid startDate
+    var removeFormErrors = function() {
+        //remove the error classes and glyphicons from the form inputs
+        removeFormError('#dateFormGroup', '#inputError-dateFormGroup');
+        removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
+        removeFormError('#titleFormGroup', '#inputError-titleFormGroup');
+        removeGlyphicon('#span-titleFormGroup');
+    };
+    
+    var validateDates = function(startDate, endDate) {
```
johntyree: Elsewhere in the code, "xxxDate" is a moment object and the string version is called "xxxDateString". You have it backwards here.

calendar_reboot.js
```diff
@@ -66,44 +65,54 @@ $(document).ready(function() {
...
+        var startDateMoment = moment(startDate, "YYYY-MM-DD");
+        var endDateMoment = moment(endDate, "YYYY-MM-DD");
         
         if (startDateMoment.isBefore(endDateMoment)) {
+            removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
             return true;
         }
         else {
-            console.log("endDate is before startDate");
+            alert("The end date needs to be after the start date.");
```
johntyree: In general, alerts are terrible.

Maybe file a bug to figure out how to do a nice error toast bar thing, e.g. http://mdbootstrap.com/javascript/alerts/

Also the widgets there are pretty sexy in general. Maybe investigate material design bootstrap?

calendar_reboot.js
```diff
@@ -66,44 +65,54 @@ $(document).ready(function() {
...
+            addFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
             return false;
         }
+        
     };
     
     var validateInput = function(formGroupId) {
+        console.log("validating the input " + formGroupId);
```
johntyree: You don't typically want these all over the place. If you have something that is not expected and always relevant then go ahead and use it, otherwise leave them out of the finished verson.

calendar_reboot.js
```diff
@@ -129,6 +138,23 @@ $(document).ready(function() {
...
     };
     
+    var validateForm = function() {
+        console.log("validating the form");
+        
+        var validateTitle = validateInput(calendarTitleId);
+        var validateStartDate = validateInput(startDateId);
+        var validateEndDate = validateInput(endDateId);
+        
+        if (validateTitle && validateStartDate && validateEndDate) {
```
johntyree:
```
isValid = validTitle && validStartDate && validEndDate
return isValid
```

calendar_reboot.js
```diff
@@ -152,6 +178,7 @@ $(document).ready(function() {
     $('#clearButton').click(function() {
         //clear all the entries in the build calendar form
         console.log("clearing form...");
```
johntyree: Again, this doesn't really tell you anything new or interesting.

calendar_reboot.js
```diff
@@ -161,17 +188,13 @@ $(document).ready(function() {
         // and fill it with the information given in the form
         //required info are startDate and title. default to tracking 1
```
johntyree: Is this still true?

calendar_reboot.js
```diff
@@ -161,17 +188,13 @@ $(document).ready(function() {
...
         var calendarTitle = $("#calendarTitle").val();
         var startDate = $("#startDate").val();
         var endDate = $("#endDate").val();
-            
         
-        var validateTitle = validateInput(calendarTitleId);
-        var validateStartDate = validateInput(startDateId);
-        var validateEndDate = validateInput(endDateId);
-        var validateDates = validateForm(startDate, endDate);
+        if (validateForm() && validateDates(startDate, endDate)) {
```
johntyree: Validating the dates is part of validating the form isn't it?

johntyree (changes requested):

calendar_reboot.js
andrayantelo: I changed it to xxxDateString. Elsewhere in the code, xxxDateString is formatted "YYYYMMDD" but here it is "YYYY-MM-DD" because this function is taking the string that is obtained from the datetimepicker. Is that a problem?

calendar_reboot.js
andrayantelo: You mean the print statement? I had it for testing purposed but forgot to delete it. :smile:

calendar_reboot.js
andrayantelo: That is certainly much prettier. :heart:

calendar_reboot.js
andrayantelo: No :smile_cat:

calendar_reboot.js
andrayantelo: Yes but I need to first check if inputs were given, then I need to check if the date inputs are valid. That's why I don't do them in the same function.

andrayantelo: I mean, if you have validateDates inside of validateForm, validateForm will first check that inputs were given for all input types (checking that the fields weren't left blank). Then check if End Date is indeed after Start Date. 

I guess this could all go in the same function. I have to think about it a little more.

calendar_reboot.js
andrayantelo: Got rid of the alert. Added help text below the end date input field that says "The end date must be after the start date." The whole input box turns red if the user provides an invalid date (a date before the start date or no input at all or null).

calendar_reboot.js
andrayantelo: @johntyree

calendar_reboot.js
andrayantelo: Deleted.

calendar_reboot.js
andrayantelo: So, I included validateDates in the validateForm function. The validateForm function first checks that input was actually given. When the user has filled out all three fields, it then checks if the dates are valid (whether end date occurs after start date). So now in the set button click handler you only have to check if validateForm is true and then you can build your calendar.

calendar_reboot.js
```diff
@@ -138,19 +135,29 @@ $(document).ready(function() {
...
         
         var validateTitle = validateInput(calendarTitleId);
         var validateStartDate = validateInput(startDateId);
         var validateEndDate = validateInput(endDateId);
         
-        if (validateTitle && validateStartDate && validateEndDate) {
-            return true;
+        var isValid = validateTitle && validateStartDate && validateEndDate;
+        
+        if (isValid) {
```
johntyree: If you're returning a bool from inside an if statement, check to see if you can just return the condition variable itself.

```
if (x) {
  return true
} else {
  return false
}
```
is the same as `return x`.

Here you could do
```
if (isValid) {
  var validDates = validateDates(...)
  return validDates
} else {
  return false
}
```
or
```
if (!isValid) {
  return false
}
return validateDates(...)
```

or even just `return (isValid && validateDates(startDateString, endDateString))`

johntyree (changes requested):

calendar_reboot.js
andrayantelo: Oh. That sure is much better. Changed it to your last suggestion.

calendar_reboot.js
```diff
@@ -66,35 +65,42 @@ $(document).ready(function() {
...
+        removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
+        removeFormError('#titleFormGroup', '#inputError-titleFormGroup');
+        removeGlyphicon('#span-titleFormGroup');
+    };
+    
+    var validateDates = function(startDateString, endDateString) {
         
-        var startDateMoment = moment(startDate, "YYYYMMDD");
-        var endDateMoment = moment(endDate, "YYYYMMDD");
+        var startDateMoment = moment(startDateString, "YYYY-MM-DD");
```
johntyree: Don't you just call these `startDate` and `endDate` elsewhere?

calendar_reboot.js
```diff
@@ -129,6 +135,21 @@ $(document).ready(function() {
...
+    var validateForm = function(startDateString, endDateString) {
+        
+        
+        var validateTitle = validateInput(calendarTitleId);
+        var validateStartDate = validateInput(startDateId);
+        var validateEndDate = validateInput(endDateId);
+        
+        var isValid = validateTitle && validateStartDate && validateEndDate;
+        
+        return (isValid && validateDates(startDateString, endDateString);
```
johntyree, line 147: Just a note that this only works because `&&` short-circuits and so `validateDates` only gets called when `isValid` is true.

johntyree (changes requested):

johntyree:
Try adding "Fixes #63" to the PR description.

andrayantelo:
Done

calendar_reboot.js
andrayantelo: Changed it to startDate and endDate

calendar_reboot.js
andrayantelo: Yes... isn't that what I want?

johntyree: Yes. I'm just calling attention to it because it's not obvious at first glance.

andrayantelo: So, is this PR ready to go?

johntyree (approved):

PR #70, opened 2017-01-09
adding calendar to dropdown after creating and saving it and before 

andrayantelo:
building it.


calendar_reboot.js
```diff
@@ -274,7 +277,7 @@ $(document).ready(function() {
           if (key !== "current_active_calendar") {
```
johntyree: unrelated, but this should be `current_active_calendar`, not `"current_active_calendar"`

johntyree (approved):

calendar_reboot.js

PR #73, opened 2017-01-10
deleteCalendar method for calendar object

andrayantelo:
also wrote removeActiveCalendar as a calendar object method. setActiveCalendar is already a method on calendar object. These methods remove the calendar from current_active_status or stores it as such respectively.

Not sure if they should be calendar object methods though.
Also, I wrote the deleteCalendar method and want to use it in the deleteButton click handler. To do that I'd need to create a calendar object. 

The first thing that happens in the deleteButton click handler is that the current_active_calendar Id is loaded from storage. So, I could use that Id, to load the calendarState from storage, and then make a calendar object using that state. THEN use the delete method using that calendar object.

Should I go about it in that way? @johntyree 

(also added var self = this in the new methods I had written)

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -248,6 +248,8 @@ $(document).ready(function() {
             var currentCalendarId = loadFromLocalStorage(current_active_calendar);
             removeFromCalendarDropdown(currentCalendarId);
             
+            
+            //below does not work
             calendar.deleteCalendar();
             calendar.removeActiveCalendar();
             
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -1,4 +1,6 @@
 var masterKey = 'masterKey';
+//the current_active_calendar is the key for localStorage that stores
+//the active calendar's Id
 var current_active_calendar = 'current_active_calendar';
 
 $(document).ready(function() {
@@ -184,6 +186,7 @@ $(document).ready(function() {
             
             //make calendar object
             var calendar = new Calendar(calendarState);
+            console.log(calendar);
             calendar.setActiveCalendar();
             calendar.saveCalendar();
             
@@ -243,15 +246,11 @@ $(document).ready(function() {
         if (confirmation === true) {
             
             var currentCalendarId = loadFromLocalStorage(current_active_calendar);
-            console.log(currentCalendarId);
-            delete calendarUniqueId[currentCalendarId];
-            storeInLocalStorage(masterKey, calendarUniqueId);
-            
             removeFromCalendarDropdown(currentCalendarId);
-            //remove the calendarstate from localStorage
-            removeFromLocalStorage(currentCalendarId);
             
-            removeFromLocalStorage(current_active_calendar);
+            calendar.deleteCalendar();
+            calendar.removeActiveCalendar();
+            
             
             console.log("clearing the page");
             //clear the page
@@ -674,14 +673,22 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
 
 
 Calendar.prototype.setActiveCalendar = function() {
-    
+    var self = this;
     //set the current active calendar 
     storeInLocalStorage(current_active_calendar, self.calendarState.uniqueId);
+};
+
+Calendar.prototype.removeActiveCalendar = function() {
+    //removes a calendar from current_active_calendar status. It will no
+    //longer be stored in LocalStorage as the current_active_calendar
+    var self = this;
+    removeFromLocalStorage(current_active_calendar);
+    
 };
     
 Calendar.prototype.saveCalendar = function() {
         //saves the calendar in localStorage, saves the calendarUniqueId object
-        //adds the calendar to the saved calendars dropdown
+        
         var self = this;
         //store the state in localStorage
         storeInLocalStorage(self.calendarState.uniqueId, self.calendarState);
@@ -690,7 +697,23 @@ Calendar.prototype.saveCalendar = function() {
         calendarUniqueId[self.calendarState.uniqueId] = self.calendarState.calendarTitle;
         storeInLocalStorage(masterKey, calendarUniqueId);
         
-    };
+};
+    
+Calendar.prototype.deleteCalendar = function() {
+    //deletes the calendar from localStorage, deletes the calendar from the
+    //calendarUniqueId object
+    
+    var self = this;
+    //delete the calendar from the calendarUniqueId object
+    delete calendarUniqueId[calendar.calendarState.uniqueId];
+    //save that change in localstorage
+    storeInLocalStorage(masterKey, calendarUniqueId);
+    
+    //remove the calendarstate from localStorage
+    removeFromLocalStorage(calendar.calendarState.uniqueId);
+
+}
+    
 
 //dictionary of savedCalendars, calendar title: unique ID
 
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -675,48 +675,6 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
     });
 };
 
-
-Calendar.prototype.setActiveCalendar = function() {
-    var self = this;
-    //set the current active calendar 
-    storeInLocalStorage(current_active_calendar, self.state.uniqueId);
-};
-
-Calendar.prototype.removeActiveCalendar = function() {
-    //removes a calendar from current_active_calendar status. It will no
-    //longer be stored in LocalStorage as the current_active_calendar
-    var self = this;
-    removeFromLocalStorage(current_active_calendar);
-    
-};
-    
-Calendar.prototype.saveCalendar = function() {
-        //saves the calendar in localStorage, saves the allCalendarIds object
-        
-        var self = this;
-        //store the state in localStorage
-        storeInLocalStorage(self.state.uniqueId, self.state);
-        
-        //put calendar in allCalendarIdss and store it
-        allCalendarIds[self.state.uniqueId] = self.state.title;
-        storeInLocalStorage(allCalendarIdsKey, allCalendarIds);
-        
-};
-    
-Calendar.prototype.deleteCalendar = function() {
-    //deletes the calendar from localStorage, deletes the calendar from the
-    //allCalendarIds object
-    
-    var self = this;
-    //delete the calendar from the allCalendarIds object
-    delete allCalendarIds[self.state.uniqueId];
-    //save that change in localstorage
-    storeInLocalStorage(allCalendarIdsKey, allCalendarIds);
-    
-    //remove the calendarstate from localStorage
-    removeFromLocalStorage(self.state.uniqueId);
-
-}
     
 
 //dictionary of savedCalendars, calendar title: unique ID
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -146,7 +146,7 @@ $(document).ready(function() {
         
         clearPage();
         buildCalendar(calendarObj);
-        store.setActive(calendarObj);
+        store.setActiveById(calendarObj.state.uniqueId);
         store.save(calendarObj);
     };
     
@@ -183,7 +183,7 @@ $(document).ready(function() {
 
             var calendar = new Calendar(state);
             
-            store.setActive(calendar);
+            store.setActiveById(calendar.state.uniqueId);
             store.save(calendar);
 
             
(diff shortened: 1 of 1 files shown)


calendar_reboot.js
```diff
@@ -243,15 +246,13 @@ $(document).ready(function() {
         if (confirmation === true) {
```
johntyree: This is just a verbose way to write `if (confirmation) {`

calendar_reboot.js
```diff
@@ -243,15 +246,13 @@ $(document).ready(function() {
...
-            delete calendarUniqueId[currentCalendarId];
-            storeInLocalStorage(masterKey, calendarUniqueId);
-            
             removeFromCalendarDropdown(currentCalendarId);
-            //remove the calendarstate from localStorage
-            removeFromLocalStorage(currentCalendarId);
             
-            removeFromLocalStorage(current_active_calendar);
+            
+            //below does not work
```
johntyree: Why not? They look like they should work to me.

calendar_reboot.js
```diff
@@ -1,4 +1,6 @@
 var masterKey = 'masterKey';
```
johntyree: It's time to pick a better name for this. How about `allCalendarIdsKey` and renaming `calendarUniqueId` to `allCalendarIds` ?

johntyree (changes requested):
It's becoming apparent that we need something to manage storage in general and that it should not be internal to the calendar object. It would be nice to have an interface like:

```
// Make a storage manager. Keep them separate using an id.
var store = new LocalCalendarStorage({store_id: appName})

if (!store.save(calendar)) { error('Unable to save calendar') }

var calendar = store.load(calendarId)
if (calendar === null) { error(Unable to load calendar) }

store.setActive(calendar)
// or
store.setActiveById(calendarId)
```

etc...

This will probably make it easier to switch to a different storage method someday and will certainly make it easier to test.

calendar_reboot.js

calendar_reboot.js
andrayantelo: :+1:

calendar_reboot.js
andrayantelo: I explained above why it doesn't work. I don't have access to the calendar object. I'd have to create a new one......

johntyree: Oh. Yes, go ahead. The calendarState is the essence of the calendar. The calendar object is just a wrapper that lets you do stuff with it. You can make and destroy it as needed. Two calendar objects made from the same state *should* be interchangeable.

andrayantelo:
@johntyree I pushed new changes to this PR. Is there a way to update the PR or is it done automatically? Anyway I made the storage manager and updated the code with the new methods of this object.

johntyree:
The PR is kept in sync with the branch on the remote, so updates show up automatically, including that conflict.

calendar_reboot.js
```diff
@@ -160,41 +157,41 @@ $(document).ready(function() {
...
-            calendar.setActiveCalendar();
-            calendar.saveCalendar();
+            var calendar = new Calendar(state);
+            
+            store.setActive(calendar);
+            store.save(calendar);
             
             //add calendar to dropdown
-            addCalendarToDropdown(calendar.calendarState.uniqueId, calendar.calendarState.calendarTitle);
+            addCalendarToDropdown(calendar.state.uniqueId, calendar.state.title);
```
johntyree, line 207: You already had the state object. No need to go through the calendar object.

calendar_reboot.js
```diff
@@ -230,29 +223,31 @@ $(document).ready(function() {
...
         //ask the user if they are sure they want to delete their calendar
         var confirmation = confirm("Are you sure you want to delete your calendar?");
-        if (confirmation === true) {
+        if (confirmation) {
             
             var currentCalendarId = loadFromLocalStorage(current_active_calendar);
             removeFromCalendarDropdown(currentCalendarId);
             
+            //load calendarState from local storage and make a calendar object
+            var calendarState = loadFromLocalStorage(currentCalendarId);
```
johntyree, line 249: store.load?

calendar_reboot.js
```diff
@@ -265,27 +260,34 @@ $(document).ready(function() {
     
     
     //WHEN PAGE LOADS
-    //first load the calendarUniqueId from storage.
-    calendarUniqueId = loadFromLocalStorage(masterKey);
+    //first load the allCalendarIds from storage.
+    allCalendarIds = store.getAllCalendarIds();
```
johntyree: `store.getAllCalendarIds() || {};`

calendar_reboot.js
```diff
@@ -265,27 +260,34 @@ $(document).ready(function() {
...
     //if it's not in localstorage make it an empty object
-    if (calendarUniqueId === null) {
-        calendarUniqueId = {};
+    if (allCalendarIds === null) {
```
johntyree: Can ditch this.

calendar_reboot.js
```diff
@@ -265,27 +260,34 @@ $(document).ready(function() {
...
     
+    //get the current_active calendar id from storage if any
+    var activeCalendarId = store.getActive();
     
-    var activeCalendar = loadFromLocalStorage(current_active_calendar);
-    
-    if (activeCalendar !== null) {
-        displayCalendar(activeCalendar);
+    if (activeCalendarId !== null) {
+        console.log("the activeCalendarKey was not null");
```
johntyree: Probably remove this. Maybe replace with a more useful message instead. `"Loading active calendar <...>"`

andrayantelo:
@johntyree Do I have to do the @ your name here or are you going to know I commented? Anyway,  not sure why I got those conflicts but I resolved them.

calendar_reboot.js
```diff
@@ -573,7 +575,7 @@ var makeCalendarObj = function(state) {
     var startDate = state.startDate;
     var numberOfYears = state.numberOfYears;
     var calendar = new Calendar(startDate, numberOfYears);
```
johntyree: This looks wrong. Does a calendar `state` still have `numberOfYears` in it?

johntyree: Also, you don't even construct a Calendar this way. You pass it a state, no?

calendar_reboot.js
```diff
@@ -619,7 +621,7 @@ var Calendar = function(state) {
     //number of months we will need to be able to cover all the years the
     //user wants to track
     self.numberOfMonths = self.endDate.diff(self.startDate, 'months', true);
-    self.title = state.calendarTitle;
+    self.title = state.title;
```
johntyree: Why have this? It's a duplicate of `state.title`. You can always just do `calendar.state.title` if you want the title.

calendar_reboot.js
```diff
@@ -689,39 +691,114 @@ Calendar.prototype.removeActiveCalendar = function() {
...
+        
+    };
+    
+    self.getAllCalendarIds = function() {
+        //gets the allCalendarIds object from storage
+        
+        return loadFromLocalStorage(toKey(allCalendarIdsKey));
+    };
+    
+    self.save = function(appObj) {
```
johntyree: Since we only have calendar objects and this is a `LocalCalendarStorage`, it's probably best to just refer to these objects as `calendar`.

Alternatively, you could call this thing a `LocalStore` or something, but I feel like it might be more abstract than it needs to be. It makes it harder to know what the functions are assuming about the incoming data.

johntyree: Feel free to do that in this PR, a different PR, or not at all.

calendar_reboot.js
```diff
@@ -689,39 +691,114 @@ Calendar.prototype.removeActiveCalendar = function() {
...
+        //remove an app object (like a calendar) from storage
+        
+        //delete the calendar from the allCalendarIds object
+        delete allCalendarIds[appObj.state.uniqueId];
+        //save that change in localstorage
+        storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
+        
+        //remove the object's state from localStorage
+        removeFromLocalStorage(toKey(appObj.state.uniqueId));
+    };
```
johntyree, line 720: What if the calendar being removed is the currently active calendar?

calendar_reboot.js
```diff
@@ -689,39 +691,114 @@ Calendar.prototype.removeActiveCalendar = function() {
...
+    
+    self.loadById = function(appObjId) {
+        //load an App object using it's Id
+        return loadFromLocalStorage(toKey(appObjId));
+    };
+    
+    self.setActive = function(appObj) {
+        //set an app object (calendar) as the current_active object/calendar
+        storeInLocalStorage(toKey(current_active_calendar), appObj.state.uniqueId);
```
johntyree: this should just call `setActiveById`

johntyree: Or maybe get rid of it. Do you use it?

calendar_reboot.js
```diff
@@ -689,39 +691,114 @@ Calendar.prototype.removeActiveCalendar = function() {
...
 //dictionary of savedCalendars, calendar title: unique ID
 
-var calendarUniqueId = {
+var allCalendarIds = {
 };
 
+//Make a storage manager
+var LocalCalendarStorage = function(params) {
+    var self = this;
+    var prefix = params['storeId'] || "";
```
johntyree, line 678: if the Id is empty, this will prefix everything with a `_` which we probably don't want.

Maybe like, `var prefix = params['storeId'] ? params['storeId'] + '_' : '';`

And then `toKey` can just be `prefix + id`.

johntyree: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

johntyree (changes requested):

calendar_reboot.js
andrayantelo: Done

calendar_reboot.js
andrayantelo: Done

calendar_reboot.js
andrayantelo: Removed

calendar_reboot.js
andrayantelo: Ah, that's an old function we don't use anymore. Deleting it.

calendar_reboot.js
andrayantelo: Removed the title from calendar object but kept it in the calendar state.

calendar_reboot.js
andrayantelo: Different PR. Maybe make an issue.

andrayantelo: So that I remember.

calendar_reboot.js
andrayantelo: Then you have to do removeActiveStatus after, which is also a storage manager method.

calendar_reboot.js
andrayantelo: I use it. But it can probably be replaced by setActiveById. Should I do that?

calendar_reboot.js
johntyree: Issue link?

calendar_reboot.js
johntyree: Probably just keep `setActiveById`, it's easy enough to get the id if you have the calendar object or the state

calendar_reboot.js
andrayantelo: I didn't make one yet lol

andrayantelo: I'm just going to change it right now.

calendar_reboot.js
johntyree: Is there ever a time when you wouldn't want to do this? Seems like something that should be handled automatically.

calendar_reboot.js
andrayantelo: No, I mean, should I replace setActive with setActiveById, and never use setActive again?

johntyree: Sure, that should be fine.

calendar_reboot.js
andrayantelo: So, shall I include removeActiveStatus inside of the remove method (which removes a calendar). Then also, in save I'd have to include the setActive.

calendar_reboot.js
```diff
@@ -226,31 +223,31 @@ $(document).ready(function() {
         //the name from the saved calendars dropdown
         
         //when you delete a calendar, it needs to be removed from local Storage
-        //it's unique Id needs to be removed from the calendarUniqueId dictionary
+        //it's unique Id needs to be removed from the allCalendarIds dictionary
```
johntyree: Its

calendar_reboot.js
```diff
@@ -226,31 +223,31 @@ $(document).ready(function() {
...
         //it's name needs to be removed from the saved calendars dropdown
```
johntyree: Its

calendar_reboot.js
```diff
@@ -226,31 +223,31 @@ $(document).ready(function() {
...
         //clear the calendar from the page and uncollapse build calendar
         
         //need to check if there is a calendar loaded that needs to be deleted.
```
johntyree: Comments that are instructions for future you should begin with `TODO:` or `TODO(@andrayantelo):` if it should be known who is responsible for doing it.

calendar_reboot.js
andrayantelo: :love_letter:

calendar_reboot.js
johntyree: Or even `TODO(#bugnum):` or something. Depending on what you're trying to say

johntyree (approved):

PR #78, opened 2017-01-12
Build calendar

### LINKED ISSUE #76: Prompt user to build a calendar immediately if no calendars exist.
When there are no saved calendars, we should just show the "build calendar" menu immediately. It's not obvious what to do, otherwise.

andrayantelo:
Shows the build calendar form when the page is loaded and there is no current active calendar. I don't really understand why so many changes are being shown... I made this branch, changed two lines, pushed, and made this PR.

Fixes #76

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -89,10 +89,10 @@ $(document).ready(function() {
     
     var validateDates = function(startDateString, endDateString) {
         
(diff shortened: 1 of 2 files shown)


johntyree:
It's showing many changes because you probably did not pull the merged features into master in your local branch.

PR #88, opened 2017-01-15
refactored the code. Created a checkit object to represent the applic…

andrayantelo:
…ation.

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -1,304 +1,322 @@
-var allCalendarIdsKey = 'allCalendarIdsKey';
-//the current_active_calendar is the key for localStorage that stores
-//the active calendar's Id
-var current_active_calendar = 'current_active_calendar';
 
-$(document).ready(function() {
+// Initializes CheckIt
+function CheckIt() {
+    // Shortcuts to DOM elements.
     
-    //will need to use a better key than this for allCalendarIdsKey (if this is
-    //even a good way to do this)
+    this.calendarTitleSelector = $('#titleFormGroup');
+    this.startDateSelector = $('#dateFormGroup');
+    this.endDateSelector = $('#dateFormGroup2');
+    this.startDatePicker = $('#datetimepicker1');
+    this.startDatePickerInput = $('#datetimepicker1 input');
+    this.endDatePicker = $('#datetimepicker2');
+    this.endDatePickerInput = $('#datetimepicker2 input');
+    this.calendarDropdown = $('#calendarDropdown');
+    this.startDateErrorSpan = $('#inputError-dateFormGroup');
+    this.endDateErrorSpan = $('#inputError-dateFormGroup2');
+    this.titleErrorSpan = $('#inputError-titleFormGroup');
+    this.titleGlyphiconTag = $('#span-titleFormGroup');
+    this.clearButton = $('#clearButton');
+    this.fullForm = $('#fulLForm');
+    this.createButton = $('#createButton');
+    this.deleteButton = $('#deleteButton');
+    this.calendarTitle = $('#calendarTitle');
+    this.startDate = $('#startDate');
+    this.endDate = $('#endDate');
+    this.buildCalendarForm = $('#collapseOne');
+    this.calendarDiv = $('#calendarDiv');
+    this.calendarTemplate = $('#template');
     
     var monthObjects;
     
-    var calendarTitleId = 'titleFormGroup'; //actual div id
-    var startDateId = 'dateFormGroup';  //also div id
-    var endDateId = 'dateFormGroup2'
+    // Initialize storage.
+    this.store = new LocalCalendarStorage({'storeId': 'checkit'})
     
-    //have the calendar show when you click in the input section of the date
+    this.clearButton.click(this.clearForm.bind(this));
+    this.createButton.click(this.createCalendar.bind(this));
+    this.calendarDropdown.on('click', 'li', this.loadFromDropdown.bind(this));
+    this.deleteButton.click(this.deleteCalendar.bind(this));
+    
+      //have the calendar show when you click in the input section of the date
     //timepicker
     
-    $('#datetimepicker1').datetimepicker({format: "YYYY-MM-DD"});
-    $('#datetimepicker1 input').click(function(event){
-        $('#datetimepicker1 ').data("DateTimePicker").show();
-    });
+    this.startDatePicker.datetimepicker({format: "YYYY-MM-DD"});
+    this.startDatePickerInput.click(function(event){
+       this.startDatePicker.data("DateTimePicker").show();
+    }.bind(this));
     
-    $('#datetimepicker2').datetimepicker({format: "YYYY-MM-DD"});
-    $('#datetimepicker2 input').click(function(event){
-        $('#datetimepicker2 ').data("DateTimePicker").show();
-    });
-    
-    //update the current active calendar
-
-    
-    var addCalendarToDropdown = function(uniqueId, title) {
-        //add the calendar with unique Id, uniqueId, and title, title, to
-        //the saved calendars dropdown on the navbar.
-        
-        
-        $('#calendarDropdown').append('<li id="' + uniqueId
-                + '"><a href=#>' + title + '</a></li>');
-    };
-    
-    var removeFromCalendarDropdown = function(uniqueId) {
-        //remove the calendar with uniqueId from the saved calendars dropdown
-        //in the navbar
-        
-        $('#calendarDropdown').children('#' + uniqueId).remove();
-    };
-    
-    
-    var addFormError = function(id, srId) {
-        //adds the classes has-error and has-feedback
-        //on the div with the class form-group and it removes the hidden
-        //class on the span with class sr-only (for screen readers to notify
-        //that there has been an error
-        
-        //id is the id of the eleent with the error glyphicon
-        //srId is the id for the span tag with the instructions for 
-        //screen readers, span has class sr-only
-        
-        $(id).addClass('has-error has-feedback');
-        $(srId).removeClass('hidden');
-        
-    };
-    
-    var removeFormError = function(id, srId) {
-        //removes the classes has-error and has-feedback from the div
-        //with the class form-group, also adds the hidden class on the span
-        //with class sr-only (for screen readers)
-        
-        $(id).removeClass('has-error has-feedback');
-        $(srId).removeClass('hidden');
-    };
-    
-    var addGlyphicon = function(id) {
-        //removes the hidden class to the glyphicon tag with id = id
-        $(id).removeClass('hidden');
-    };
-    
-    var removeGlyphicon = function(id) {
-        //adds the hidden class to the glyphicon tag with id=id
-        $(id).addClass('hidden');
-    };
-    
-    var removeFormErrors = function() {
-        //remove the error classes and glyphicons from the form inputs
-        removeFormError('#dateFormGroup', '#inputError-dateFormGroup');
-        removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
-        removeFormError('#titleFormGroup', '#inputError-titleFormGroup');
-        removeGlyphicon('#span-titleFormGroup');
-    };
-    
-    var validateDates = function(startDateString, endDateString) {
-        
-        var startDate = moment(startDateString, "YYYY-MM-DD");
-        var endDate = moment(endDateString, "YYYY-MM-DD");
-        
-        if (startDate.isBefore(endDate)) {
-            removeFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
-            return true;
-        }
-        else {
-            addFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
-            return false;
-        }
-        
-    };
-    
-    var validateInput = function(formGroupId) {
-        
-        //make a variable for the id string
-        var id = formGroupId;
-        //a variable for the id string with the # infront of it
-        var $id = $('#' + formGroupId);
-   //     //a variable for the input tag value
-        var inputValue = $id.find('input[type=text]').val();
-        
-        //if the user hasn't written anything in the input tag
-        // then we need to bring up the error, the glyphicon needs
-        //to appear
-        if (inputValue === "" || inputValue === null) {
-            addFormError($id, $('#inputError-' + id) )
-            
-            //reveal the error glyphicon, ONLY THE TITLE HAS A GLYPHICON
-            //check if we are dealing with the title
-            if ($id.find("input").attr("id") === "calendarTitle") {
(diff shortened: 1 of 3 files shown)


calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
 
-$(document).ready(function() {
-    
-    //will need to use a better key than this for allCalendarIdsKey (if this is
-    //even a good way to do this)
+// Initializes CheckIt
+function CheckIt() {
+    // Shortcuts to DOM elements.
+    
+    this.calendarTitleSelector = $('#titleFormGroup');
```
johntyree: These are *all* selectors right? Probably should rename them like `this.$calendarTitle`.

johntyree: :smiling_imp:

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-                + '"><a href=#>' + title + '</a></li>');
-    };
-    
-    var removeFromCalendarDropdown = function(uniqueId) {
-        //remove the calendar with uniqueId from the saved calendars dropdown
-        //in the navbar
+        if (activeCalendarState !== null) {
+            var state = activeCalendarState;
+            var calendar = new Calendar(state, this);
+            console.log(calendar);
```
johntyree: Remove

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-    var addFormError = function(id, srId) {
-        //adds the classes has-error and has-feedback
-        //on the div with the class form-group and it removes the hidden
-        //class on the span with class sr-only (for screen readers to notify
-        //that there has been an error
+}
+
+CheckIt.prototype.clearForm = function() {
+    // Clears the buildCalendar form.
+    removeFormErrors();
```
johntyree: `this.removeFormErrors()` ?

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+        //load calendarState from local storage and make a calendar object
+        var calendarState = this.store.loadById(currentCalendarId);
         
-        var validateTitle = validateInput(calendarTitleId);
-        var validateStartDate = validateInput(startDateId);
-        var validateEndDate = validateInput(endDateId);
+        var calendar = new Calendar(calendarState, this);
         
-        var isValid = validateTitle && validateStartDate && validateEndDate;
+        //delete the calendar and remove it's active calendar status
```
johntyree: its

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+    var title = this.calendarTitle.val();
+    var start = this.startDate.val();
+    var end = this.endDate.val();
+    
+    if (this.validateForm(start, end)) {
         
-        //id is the id of the eleent with the error glyphicon
-        //srId is the id for the span tag with the instructions for 
-        //screen readers, span has class sr-only
+        //clear the previously displayed calendar <-- or should i reload the page???
```
johntyree: Definitely don't reload the page. Just clear your div. What else is there to worry about?

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+        this.buildCalendar(calendar);
+        this.buildCalendarForm.collapse('hide'); 
+   }
+};
+
+CheckIt.prototype.loadFromDropdown = function( event ) {
+    //when a calendar is clicked on in the dropdown menu
+    //the calendar that is currently on display (if there is one)
+    //needs to be removed and replaced with the calendar that was clicked on
+    //from the dropdown menu. Its' data will be loaded from localstorage
```
johntyree: Drop this middle sentence. It adds nothing

> from the dropdown menu. Load the data, clear the page, build the saved calendar.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-        else {
-            addFormError('#dateFormGroup2', '#inputError-dateFormGroup2');
-            return false;
-        }
-        
-    };
+    var calendar = new Calendar(state, this);
+    
+    this.displayCalendar(calendar);
+    this.buildCalendarForm.collapse('hide');
```
johntyree: `this.hideBuildMenu` and `this.showBuildMenu` would be nice.

johntyree: That way, when you decide that a sliding form that never goes away isn't a good way to do this, you'll only have to change those two functions.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-    };
+    var calendar = new Calendar(state, this);
+    
+    this.displayCalendar(calendar);
+    this.buildCalendarForm.collapse('hide'); 
+    
+};
+
+CheckIt.prototype.deleteCalendar = function() {
+    //deletes the current calendar on display, removes
```
johntyree: The order of these comments is misleading

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-            }
-        }
-        return true;
-    };
-    
-    var validateForm = function(startDateString, endDateString) {
+        this.removeFromCalendarDropdown(currentCalendarId);
         
+        //load calendarState from local storage and make a calendar object
+        var calendarState = this.store.loadById(currentCalendarId);
```
johntyree: Seems like we should have a `removeById`

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-        calendarObject.generateEmptyCalendar(calendarObject.monthObjects);
-        calendarObject.fillCalendar(calendarObject.monthObjects);
-        
-    };
+    this.calendarDropdown.children('#' + uniqueId).remove();
+};
+
+CheckIt.prototype.addFormError = function(id, srId) {
+
+    //adds the classes has-error and has-feedback
```
johntyree: You usually want comments to be higher level than this unless they are explaining something that isn't obvious from the code.

For example: `// Set the appropriate CSS on the form fields to indicate an error state`

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-            //clear the previously displayed calendar <-- or should i reload the page???
-            clearPage();
+CheckIt.prototype.removeFormError = function(id, srId) {
 
-            //make a calendar State
-            var state = emptyCalendarState({startDate: start, endDate: end, calendarTitle: title});
-            
-            //make calendar object
+    //removes the classes has-error and has-feedback from the div
+    //with the class form-group, also adds the hidden class on the span
```
johntyree: See above comment on CSS

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
 
-            
-            //add calendar to dropdown
-            addCalendarToDropdown(calendar.state.uniqueId, calendar.state.title);
+    //removes the hidden class to the glyphicon tag with id = id
+    id.removeClass('hidden');
+};
+
+CheckIt.prototype.removeGlyphicon = function(id) {
+    //adds the hidden class to the glyphicon tag with id=id
```
johntyree: This comment doesn't really tell you anything either.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-            var calendar = new Calendar(state);
-            
-            store.setActiveById(calendar.state.uniqueId);
-            store.save(calendar);
+CheckIt.prototype.addGlyphicon = function(id) {
 
-            
-            //add calendar to dropdown
-            addCalendarToDropdown(calendar.state.uniqueId, calendar.state.title);
+    //removes the hidden class to the glyphicon tag with id = id
```
johntyree: This comment doesn't really tell you anything that isn't obvious from the next line.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-            
-            //delete the calendar and remove it's active calendar status
-            store.remove(calendar);
-            
-            //console.log("clearing the page");
-            //clear the page
-            clearPage();
-            $('#collapseOne').collapse('show'); 
-        
+CheckIt.prototype.validateInput = function(formGroupId) {
```
johntyree: `$formGroup` instead of `formGroupId` probably?

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+    //a variable for the input tag value
+    var id = formGroupId.attr('id');
+    var inputValue = formGroupId.find('input[type=text]').val();
+    
+    //if the user hasn't written anything in the input tag
+    // then we need to bring up the error, the glyphicon needs
```
johntyree: Not to harp on comments everywhere, but this one is a good example of a comment that is likely to become wrong in the future. You'll decide you want errors to look or be handled differently and this will be out of date. It won't be updated because this code has nothing to do with how errors are shown so you won't think to go update the comments in it.

I'd say something like, `// If the user hasn't written anything we fail immediately`

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
-            store.remove(calendar);
-            
-            //console.log("clearing the page");
-            //clear the page
-            clearPage();
-            $('#collapseOne').collapse('show'); 
-        
+CheckIt.prototype.validateInput = function(formGroupId) {
+    //a variable for the input tag value
+    var id = formGroupId.attr('id');
```
johntyree: Probably useful to grab the `$input` selector here. You recreate it several times below.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+    
+    //if the user hasn't written anything in the input tag
+    // then we need to bring up the error, the glyphicon needs
+    //to appear
+    if (inputValue === "" || inputValue === null) {
+        this.addFormError(formGroupId, $('#inputError-' + id) )
+        
+        //reveal the error glyphicon, ONLY THE TITLE HAS A GLYPHICON
+        //check if we are dealing with the title
+        if (formGroupId.find("input").attr("id") === "calendarTitle") {
```
johntyree: Above you're using `.find('input[type=text]')`. Do they always return the same thing? If so, pick one.

calendar_reboot.js
```diff
@@ -1,304 +1,322 @@
...
+
+CheckIt.prototype.validateForm = function(startDateString, endDateString) {
         
-    });
+        
+    var validateTitle = this.validateInput(this.calendarTitleSelector);
+    var validateStartDate = this.validateInput(this.startDateSelector);
+    var validateEndDate = this.validateInput(this.endDateSelector);
     
+    var isValid = validateTitle && validateStartDate && validateEndDate;
```
johntyree, line 274: I'd probably add a comment about checking that the dates exist before trying to validate them. I.e. `isValid` should pass first, otherwise `this.validateDates` will probably explode.

Maybe `// Make sure input is OK before parsing and validating dates`

calendar_reboot.js
```diff
@@ -679,7 +691,11 @@ var allCalendarIds = {
...
     var self = this;
     var prefix = params['storeId'] || "";
-    
+    var allCalendarIdsKey = 'allCalendarIdsKey';
+    //the current_active_calendar is the key for localStorage that stores
+    //the active calendar's Id
+    var current_active_calendar = 'current_active_calendar';
+
 
     //params.storeId
```
johntyree: Remove

johntyree (changes requested):

calendar_reboot.js
andrayantelo: :scream:

calendar_reboot.js
andrayantelo: Removed.

calendar_reboot.js
andrayantelo: Fixed.

calendar_reboot.js
andrayantelo: ok.

calendar_reboot.js
andrayantelo: Ok.

calendar_reboot.js
andrayantelo: Okay. done :+1:

calendar_reboot.js
andrayantelo: Rewrote it.

calendar_reboot.js
andrayantelo: Why? I use current_active_calendar

calendar_reboot.js
```diff
@@ -3,50 +3,50 @@
...
+    this.$startDatePickerInput = $('#datetimepicker1 input');
+    this.$endDatePicker = $('#datetimepicker2');
+    this.$endDatePickerInput = $('#datetimepicker2 input');
+    this.$calendarDropdown = $('#calendarDropdown');
+    this.$startDateErrorSpan = $('#inputError-dateFormGroup');
+    this.$endDateErrorSpan = $('#inputError-dateFormGroup2');
+    this.$titleErrorSpan = $('#inputError-titleFormGroup');
+    this.$titleGlyphiconTag = $('#span-titleFormGroup');
+    this.$clearButton = $('#clearButton');
+    this.$fullForm = $('#fulLForm');
```
johntyree: Is `'#fulLForm` a typo?

johntyree (changes requested):

calendar_reboot.js
andrayantelo: Yes.

calendar_reboot.js
andrayantelo: ?

andrayantelo: @johntyree

calendar_reboot.js
johntyree: Did this happen?

calendar_reboot.js
```diff
@@ -1,311 +1,324 @@
...
     
-    var validateForm = function(startDateString, endDateString) {
+};
+
+CheckIt.prototype.deleteCalendar = function() {
+    // Removes the name from the saved calendars dropdown of the current
+    // active calendar then deletes the current calendar from storage and
+    // clears the page.
+    
+    //ask the user if they are sure they want to delete their calendar
```
johntyree: This comment would do better to explain why we're asking. It's clear from the next line what we're asking.

Or just remove it.

johntyree: e.g. `// Guard against accidental clicks of the delete button`

johntyree (changes requested):

calendar_reboot.js
andrayantelo: Yeah, I got rid of it.

calendar_reboot.js

calendar_reboot.js
johntyree: The comment

calendar_reboot.js
johntyree: No, you got rid of the sentence that says what this function does and kept the redundant and soon to be wrong one that says "data will be loaded from local storage where it was stored"

johntyree (changes requested):

johntyree (approved):

PR #90, opened 2017-01-15
Date are invalid if they are more than 5 years apart.

### LINKED ISSUE #89: Validate that the span of the calendar is under 10 - 20 years.
Have to see at what point it starts to get too slow.

andrayantelo:
Also fixed the allCalendarsId thing which you just watched me do. Fixes #89

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -314,7 +314,7 @@ CheckIt.prototype.clearPage = function() {
 
 $(document).ready(function() {
     
-    checkit = new CheckIt();
+    var checkit = new CheckIt();
     
 });
 
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -49,6 +49,7 @@ function CheckIt() {
         this.$endDatePicker.data("DateTimePicker").show();
     }.bind(this));
    
+   
     //WHEN PAGE LOADS
     //first load the allCalendarIds from storage.
     var allCalendarIds = this.store.getAllCalendarIds() || {};
@@ -59,7 +60,7 @@ function CheckIt() {
         this.addCalendarToDropdown(key, allCalendarIds[key]);
       }
     }
-    
+   
     //get the current_active calendar id from storage if any
     var activeCalendarId = this.store.getActive();
     
@@ -75,9 +76,9 @@ function CheckIt() {
         }
         
     }
-    else {
-        this.showBuildMenu();
-    }
+    //else {
+    //    this.showBuildMenu();
+    //}
     
     
 }
@@ -140,7 +141,7 @@ CheckIt.prototype.loadFromDropdown = function( event ) {
     
     var dropdownItemId = event.currentTarget.id;
     
-    var state = store.loadById(dropdownItemId);
+    var state = this.store.loadById(dropdownItemId);
     
     var calendar = new Calendar(state, this);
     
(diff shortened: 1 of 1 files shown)


calendar_reboot.js
```diff
@@ -75,9 +76,9 @@ function CheckIt() {
         }
         
     }
-    else {
-        this.showBuildMenu();
-    }
+    //else {
```
johntyree: Why is this commented out?

calendar_reboot.js
```diff
@@ -225,13 +226,23 @@ CheckIt.prototype.removeFormErrors = function() {
...
     var startDate = moment(startDateString, "YYYY-MM-DD");
     var endDate = moment(endDateString, "YYYY-MM-DD");
     
+    var difference = endDate.diff(startDate.format("YYYY-MM-DD"), 'years', true);
+    
     if (startDate.isBefore(endDate)) {
         this.removeFormError(this.$endDateForm, this.$endDateErrorSpan);
-        return true;
+        // If there are more than 5 years between the dates return false for invalid
+        if (difference > 5) {
```
johntyree: You don't need an if statement here. Just return the appropriate value.

calendar_reboot.js
```diff
@@ -703,7 +707,9 @@ var LocalCalendarStorage = function(params) {
         storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
         
         //put calendar in allCalendarIdss and store it
+        var allCalendarIds = self.getAllCalendarIds();
         allCalendarIds[calendarObj.state.uniqueId] = calendarObj.state.title;
+        console.log(allCalendarIds);
```
johntyree: remove console.log

johntyree (changes requested):

calendar_reboot.js
andrayantelo: I was checking to see where in the code allCalendarIds was getting overwritten.

andrayantelo: I uncommented it.

calendar_reboot.js
andrayantelo: Fixed.

calendar_reboot.js
andrayantelo: Removed.

johntyree:
This branch has #88 in it, so don't merge this until #88 is merged. That's why the diff is so big.

If you want to fix that, then after you merge #88 into master, merge master back in to this branch.

johntyree (approved):

calendar_reboot.js
```diff
@@ -308,7 +314,7 @@ CheckIt.prototype.clearPage = function() {
 
 $(document).ready(function() {
     
-    var checkit = new CheckIt();
+    checkit = new CheckIt();
```
johntyree: `var` needed

calendar_reboot.js
```diff
@@ -695,7 +694,9 @@ var LocalCalendarStorage = function(params) {
         storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
         
         //put calendar in allCalendarIdss and store it
+        var allCalendarIds = self.getAllCalendarIds();
```
johntyree, line 697: Nothing to change here, but in general this pattern is known as "read-modify-write" and it only works because we *know* that we are the only ones that are modifying the value.

If there were other code that could be modifying this, then you could end up "racing" where both of you read, then one of you modifies and writes but it gets "forgotten" when the second one writes because the version that the second one read didn't include the changes from the first.

johntyree (changes requested):

johntyree (approved):

PR #91, opened 2017-01-16
access the allCalendarIds object from storage before deleting an item…

andrayantelo:
… from it

### DIFF
--- a/calendar_reboot.js
+++ b/calendar_reboot.js
@@ -702,19 +702,8 @@ var LocalCalendarStorage = function(params) {
     
     self.remove = function(calendarObj) {
         //remove an app object (like a calendar) from storage
-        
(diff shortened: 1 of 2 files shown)


calendar_reboot.js
```diff
@@ -703,6 +703,8 @@ var LocalCalendarStorage = function(params) {
     self.remove = function(calendarObj) {
         //remove an app object (like a calendar) from storage
```
johntyree: https://en.wikipedia.org/wiki/Don't_repeat_yourself

`self.remove = function(calendarObj) { return self.removeById(calendarObj.state.uniqueId); }`

johntyree (changes requested):

johntyree (approved):

PR #93, opened 2017-01-22
Made the storage object work asynchronously.

andrayantelo:


public/calendar_reboot.js
```diff
@@ -677,14 +700,26 @@ var LocalCalendarStorage = function(params) {
     var toKey = function(id) {
         //make a key out of a uniqueId
         
-        return prefix + "_" + id;
-        
+        var key = prefix + "_" + id;
+        console.log("toKey("+id+") -> " + key);
```
johntyree: You can remove this log now

public/calendar_reboot.js
```diff
@@ -677,14 +700,26 @@ var LocalCalendarStorage = function(params) {
...
-        return loadFromLocalStorage(toKey(allCalendarIdsKey)) || {};
+        return new Promise( function(resolve, reject) {
+            var allCalendarIds = loadFromLocalStorage(toKey(allCalendarIdsKey));
+            
+            if (allCalendarIds !== null ) {
+                
+                resolve(allCalendarIds);
+            }
+            else {
+                console.log(allCalendarIds);
```
johntyree: You can remove this too. We have already verified that it is `null`.

public/calendar_reboot.js
```diff
@@ -694,10 +729,17 @@ var LocalCalendarStorage = function(params) {
         storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
```
johntyree: This is independent of the `getAllCalendarIds` part and should be wrapped in its own promise. Then you can compose the two together using the `Promise.*` functions and return *that*. The `save()` call isn't really "complete" until both the `state` is stored and the `allCalendarIds` are updated.

public/calendar_reboot.js
```diff
@@ -710,38 +752,60 @@ var LocalCalendarStorage = function(params) {
...
+                // Delete the calendar from allCalendarIds.
+                delete allCalendarIds[uniqueId];
+                // Save that change
+                storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
+                // Remove the calendar from local storage
+                removeFromLocalStorage(toKey(uniqueId));
+                // Remove active status from the calendar
+                removeFromLocalStorage(toKey(current_active_calendar));
+            }.bind(this))
+            .catch(function () {
```
johntyree, line 766: Is this meant to imply that no matter what happens (i.e. network problems, missing data, etc) you want to ignore it entirely without telling anyone?

public/calendar_reboot.js
johntyree: When we switch to firebase, I assume it will be possible to make several changes to the database together in a single call, but this is good practice now.

public/calendar_reboot.js
```diff
@@ -50,35 +50,52 @@ function CheckIt() {
...
+            for (var key in allCalendarIds) {
+                if (allCalendarIds.hasOwnProperty(key)) {
+                this.addCalendarToDropdown(key, allCalendarIds[key]);
+                }
+            }
+        }.bind(this))
+        
+        .catch(function (value) {
+            console.log("No calendars in storage.");
+            return;
```
johntyree: `return` is unneeded here.

public/calendar_reboot.js
```diff
@@ -50,35 +50,52 @@ function CheckIt() {
...
+    // WHEN PAGE LOADS
+    // Load the calendar Ids from storage and fill the dropdown with calendar
+    // titles.
+    this.store.getAllCalendarIds()
+        .then(function (allCalendarIds) {
+            // Add calendar titles to dropdown.
+            
+            for (var key in allCalendarIds) {
+                if (allCalendarIds.hasOwnProperty(key)) {
+                this.addCalendarToDropdown(key, allCalendarIds[key]);
```
johntyree: indent

public/calendar_reboot.js
```diff
@@ -50,35 +50,52 @@ function CheckIt() {
...
+                }
+            }
+        }.bind(this))
+
```
johntyree: It would be best not to put whitespace randomly everywhere because it tricks the reader into thinking things are unrelated. In this case, `.catch()` is being called on the return value of `.then()`, but it kind of looks like it stands alone somehow.

public/calendar_reboot.js
```diff
@@ -50,35 +50,52 @@ function CheckIt() {
...
+               .then(function (activeCalendarState) {
+                   if (activeCalendarState !==  null) {
+                       var state = activeCalendarState;
+                       var calendar = new Calendar(state, this);
+                       this.displayCalendar(calendar);
+                   }
+               }.bind(this))
+               
+               .catch(function() {
+                   this.showBuildMenu();
```
johntyree, line 89: If we got here then that means that our `activeCalendar` doesn't exist. This is a serious error and means the data is corrupted somehow. You should at a minimum `console.error()` it and maybe even try to correct the data.

public/calendar_reboot.js
```diff
@@ -136,12 +153,16 @@ CheckIt.prototype.loadFromDropdown = function( event ) {
...
-    var calendar = new Calendar(state, this);
-    
-    this.displayCalendar(calendar);
-    this.hideBuildMenu(); 
+    return this.store.loadById(dropdownItemId)
+        .then(function(state) {
+            var calendar = new Calendar(state, this);
+            this.displayCalendar(calendar);
+            this.hideBuildMenu(); 
+            }.bind(this))
```
johntyree: indentation

johntyree: also whitespace

public/calendar_reboot.js
```diff
@@ -154,18 +175,20 @@ CheckIt.prototype.deleteCalendar = function() {
...
+        var currentCalendarId = this.store.getActive()
+            .then(function(currentCalendarId) {
+                this.removeFromCalendarDropdown(currentCalendarId);
+                //delete the calendar and remove its active calendar status
+                this.store.removeById(currentCalendarId);
+                //console.log("clearing the page");
+                //clear the page
+                this.clearPage();
+                this.showBuildMenu(); 
+                }.bind(this))
```
johntyree: indentation and ws

johntyree (changes requested):

public/calendar_reboot.js
andrayantelo: No, it just means I forgot to type something in there.

public/calendar_reboot.js
```diff
@@ -710,38 +752,69 @@ var LocalCalendarStorage = function(params) {
...
+                delete allCalendarIds[uniqueId];
+                // Save that change
+                storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
+                // Remove the calendar from local storage
+                removeFromLocalStorage(toKey(uniqueId));
+                // Remove active status from the calendar
+                removeFromLocalStorage(toKey(current_active_calendar));
+            }.bind(this))
+            .catch(function () {
+                console.log("Unable to remove calendar);
```
johntyree: Missing ending quote

johntyree (changes requested):

johntyree (approved):

PR #106, opened 2017-01-23
Fixed sign in and sign out buttons CSS

andrayantelo:
Also, made the sign in function use sign in with redirect instead of with a popup.


.firebaserc
```diff
@@ -0,0 +1,6 @@
+{
+  "projects": {
+    "default": "check-it-14d35",
+    "checkit": "check-it-14d35"
+  }
+}
```
johntyree: File should end with a newline.

database.rules.json
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
+    ".read": "auth != null",
```
johntyree: This is OK for now. You'll want to further limit this later so that users can only read / write their own data.

public/calendar.css
```diff
@@ -6,6 +6,63 @@ body{
...
+     -moz-user-select: none;
+      -ms-user-select: none;
+          user-select: none;
+  background-image: none;
+  border: 1px solid transparent;
+  border-radius: 4px;
+}
+
+#user-container {
+  display: flex;
```
johntyree: My understanding is that this isn't cross platform yet and that you have to use the -moz and -webkit versions in addition to this. Maybe look into that?

.firebaserc
johntyree: missing newline

database.rules.json
johntyree, line 3: You'll want to limit this such that users can only read/write their own data eventually. (open an issue?)

johntyree (changes requested):

public/calendar_reboot.js
```diff
@@ -94,10 +109,81 @@ function CheckIt() {
        .catch(function () {
            console.log("There is no current active calendar");
            this.showBuildMenu();
-       }.bind(this));
+       }.bind(this));
```
johntyree: There is whitespace (ws) at the end of this line

public/calendar_reboot.js
```diff
@@ -94,10 +108,81 @@ function CheckIt() {
...
+        this.$userPic.removeAttr('hidden');
+        this.$signOutButton.removeAttr('hidden');
+    
+        // Hide sign-in button.
+        this.$signInButton.attr('hidden', 'true');
+    
+      }
+    else { // User is signed out!
+        // Remove the user from the store so that we can't access their information.
+        this.store.user = null;
```
johntyree, line 187: What about the active calendar and the list of calendars in the drop down?

public/index.html
```diff
@@ -90,18 +111,33 @@
...
-            
-            
+
+           </ul> <!-- /.nav navbar-nav -->  
+         <!-- third item in navbar, sign in button -->
+           <ul>
+            <div id="user-container">
+                <div hidden id="user-pic"></div>
+                <div hidden id="user-name"></div>
+                <button class="btn btn-primary navbar-btn " hidden id="sign-out"><!-- mdl-js-button mdl-js-ripple-effect mdl-color-text--white"-->
```
johntyree: What is commented out here? Should it be uncommented or should it be removed?

johntyree (changes requested):

public/index.html
andrayantelo: Should be removed.

database.rules.json
andrayantelo: Opened issue #107

public/calendar_reboot.js
johntyree: If the user has signed out, they should no longer be available.

johntyree (changes requested):

public/index.html
```diff
@@ -90,27 +114,45 @@
...
+           </ul> <!-- /.nav navbar-nav -->  
+         <!-- third item in navbar, sign in button -->
+           <ul>
+            <div id="user-container">
+                <div hidden id="user-pic"></div>
+                <div hidden id="user-name"></div>
+                <button class="btn btn-primary navbar-btn " hidden id="sign-out">
+                  Sign-out
+                </button>
+                <button class="btn btn-primary navbar-btn " hidden id="sign-in">
```
johntyree, line 127: Why does the sign-in button start out hidden?

johntyree: Apparently, `firebase.auth` always chooses one of the two buttons to unhide.

What about the other buttons?

johntyree (approved):

Issue #94, opened 2017-01-23
Give visual indication that we are waiting for storage (loading/saving)


andrayantelo:
Need to trigger an event when storage is accessed (and when it is done doing it's work) so that the storage object methods startWork and endWork will be triggered which will increment/decrement a counter respectively. The loading wheel will appear when the counter is a nonzero number. The counter indicates how many storage processes are running.

andrayantelo:
What is polyfill for IE 9 for CustomEvent object?

andrayantelo:
The eventListener will listen for the backgroundActivityChange event and run the code to display a loading wheel. If the backgroundActivityChange event is dispatched, that means the storage is being accessed.

PR #108, opened 2017-01-24
Loading wheel when storage is being accessed 

andrayantelo:
Added a customEvent object on CheckIt. Pass it to
the storage object. Added startWork and endWork methods
in storage object.
Resolves issue #94

### DIFF
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -33,6 +33,8 @@ function CheckIt() {
     this.$calendarDiv = $('#calendarDiv');
     //this.$calendarTemplate = $('#template');
     
+    this.spinner = new Spinner().spin()
+    
     var monthObjects;
     
     // Initialize storage.
@@ -71,15 +73,16 @@ function CheckIt() {
 
 CheckIt.prototype.displayLoadingWheel = function() {
     // Displays the loading wheel.
-    var spinner = new Spinner().spin()
-     
+
     // Hardcoded for now.
+    this.$buildFormAccordion.attr('hidden', true);
     this.clearPage();
-    document.getElementById("loadingWheel").appendChild(spinner.el);
+    document.getElementById("loadingWheel").appendChild(this.spinner.el);
 };
 
 CheckIt.prototype.hideLoadingWheel = function() {
-    spinner.stop();
+    this.$buildFormAccordion.removeAttr('hidden');
+    this.spinner.stop();
 };
 
 CheckIt.prototype.fillDropdown = function() {
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -33,6 +33,7 @@ function CheckIt() {
     this.$calendarDiv = $('#calendarDiv');
     //this.$calendarTemplate = $('#template');
     
+    
     var opts = {
       lines: 13 // The number of lines to draw
     , length: 28 // The length of each line
@@ -56,7 +57,7 @@ function CheckIt() {
     , position: 'absolute' // Element positioning
     }
     
-    this.spinner = new Spinner(ops);
+    this.spinner = new Spinner(opts);
     
     var monthObjects;
     
@@ -96,16 +97,14 @@ function CheckIt() {
 
 CheckIt.prototype.displayLoadingWheel = function() {
     // Displays the loading wheel.
-    var target = document.getElementById("loadingWheel");
-    // Hardcoded for now.
-    this.$buildFormAccordion.attr('hidden', true);
-    this.clearPage();
+    
+    // Needs a plugin to use jQuery.
+    var target = document.getElementById('loadingWheel');
     this.spinner.spin(target);
     
 };
 
 CheckIt.prototype.hideLoadingWheel = function() {
-    this.$buildFormAccordion.removeAttr('hidden');
     this.spinner.stop();
 };
 
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -200,6 +200,14 @@ CheckIt.prototype.signOut = function() {
     this.auth.signOut();
 };
 
+// Triggers when there is a change in the storage.
+CheckIt.prototype.onActivityChanged = function(storageObj) {
+    // Will Manipulate the DOM to show the loading wheel or to hide it.
+    // Passing storageObj as argument to have access to activity calls, which
+    // will tell us whether or not the loadingWheel should be on display or not.
+    
+};
+
 // Triggers when the auth state change for instance when the user signs-in or signs-out.
 CheckIt.prototype.onAuthStateChanged = function(user) {
     
@@ -878,6 +886,11 @@ var LocalCalendarStorage = function(params) {
         return key;
     };
     
+    self.onActivityChanged = function(func) {
+        // Will run checkit's onActivityChanged.
+        func();
+    };
+    
 
     self.startWork = function(processName) {
         // Will increment the counter and possibly fire an event.
@@ -885,9 +898,7 @@ var LocalCalendarStorage = function(params) {
         self.activeCalls += 1;
         
         // Will dispatch the event backgroundActivityChange 
-        if (self.activeCalls > 0) {
-            console.log("dipatching backgroundActivityChange");
-        }
+        
     };
     
     self.endWork = function(accesor) {
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -33,7 +33,30 @@ function CheckIt() {
     this.$calendarDiv = $('#calendarDiv');
     //this.$calendarTemplate = $('#template');
     
-    this.spinner = null;
+    var opts = {
+      lines: 13 // The number of lines to draw
+    , length: 28 // The length of each line
+    , width: 14 // The line thickness
+    , radius: 42 // The radius of the inner circle
+    , scale: 1 // Scales overall size of the spinner
+    , corners: 1 // Corner roundness (0..1)
+    , color: '#000' // #rgb or #rrggbb or array of colors
+    , opacity: 0.25 // Opacity of the lines
+    , rotate: 0 // The rotation offset
+    , direction: 1 // 1: clockwise, -1: counterclockwise
+    , speed: 1 // Rounds per second
+    , trail: 60 // Afterglow percentage
+    , fps: 20 // Frames per second when using setTimeout() as a fallback for CSS
+    , zIndex: 2e9 // The z-index (defaults to 2000000000)
+    , className: 'spinner' // The CSS class to assign to the spinner
+    , top: '50%' // Top position relative to parent
+    , left: '50%' // Left position relative to parent
+    , shadow: false // Whether to render a shadow
+    , hwaccel: false // Whether to use hardware acceleration
+    , position: 'absolute' // Element positioning
+    }
+    
+    this.spinner = new Spinner(ops);
     
     var monthObjects;
     
@@ -69,37 +92,6 @@ function CheckIt() {
     // WHEN PAGE LOADS
     this.initFirebase();
 
-};
-
-CheckIt.prototype.setUpLoadingWheel = function() {
-    // Creates spinner object and appends it to document
-    
-    var opts = {
-      lines: 13 // The number of lines to draw
-    , length: 28 // The length of each line
-    , width: 14 // The line thickness
-    , radius: 42 // The radius of the inner circle
-    , scale: 1 // Scales overall size of the spinner
-    , corners: 1 // Corner roundness (0..1)
-    , color: '#000' // #rgb or #rrggbb or array of colors
-    , opacity: 0.25 // Opacity of the lines
-    , rotate: 0 // The rotation offset
-    , direction: 1 // 1: clockwise, -1: counterclockwise
-    , speed: 1 // Rounds per second
-    , trail: 60 // Afterglow percentage
-    , fps: 20 // Frames per second when using setTimeout() as a fallback for CSS
-    , zIndex: 2e9 // The z-index (defaults to 2000000000)
-    , className: 'spinner' // The CSS class to assign to the spinner
-    , top: '50%' // Top position relative to parent
-    , left: '50%' // Left position relative to parent
-    , shadow: false // Whether to render a shadow
-    , hwaccel: false // Whether to use hardware acceleration
-    , position: 'absolute' // Element positioning
-    }
-    
-    // If you don't include the opts as an argument the loading wheel
-    // is smaller.
-    this.spinner = new Spinner(opts);
 };
 
 CheckIt.prototype.displayLoadingWheel = function() {
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -95,11 +95,15 @@ function CheckIt() {
 
 };
 
-CheckIt.prototype.displayLoadingWheel = function() {
+CheckIt.prototype.displayLoadingWheel = function(elementId) {
     // Displays the loading wheel.
     
+    // Parameters: elementId 
+                       // A string pertaining to the id of the element
+                       // where you want to place the loading wheel.
+    
     // Needs a plugin to use jQuery.
-    var target = document.getElementById('loadingWheel');
+    var target = document.getElementById(elementId);
     this.spinner.spin(target);
     
 };
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -35,7 +35,7 @@ function CheckIt() {
     
     var monthObjects;
     
-    this.backgroundActivityChange = new customEvent('backgroundActivityChange');
+    this.backgroundActivityChange = new CustomEvent('backgroundActivityChange');
     
     // Initialize storage.
 
@@ -811,7 +811,7 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
 var LocalCalendarStorage = function(params) {
     var self = this;
     var prefix = params['storeId'] || "";
-    var event = params{'event'};
+    var event = params['event'];
     var allCalendarIdsKey = 'allCalendarIdsKey';
     //the current_active_calendar is the key for localStorage that stores
     //the active calendar's Id
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -201,7 +201,6 @@ CheckIt.prototype.onActivityChanged = function(activeCalls) {
         this.displayLoadingWheel("loadingWheel");
     }
     else if (activeCalls === 0) {
-        console.log("activity call is equal to zero!");
         this.hideLoadingWheel();
     }
     
@@ -898,7 +897,7 @@ var LocalCalendarStorage = function(params) {
         // Will increment the counter and possibly fire an event.
         
         self.activeCalls += 1;
-        console.log("store's startWork is running, activeCalls = " + self.activeCalls);
+        self.activeCalls += 1;
         
         // Will dispatch the event backgroundActivityChange 
         self.activityChangeFunctions.forEach(function(func) {
@@ -949,7 +948,6 @@ var LocalCalendarStorage = function(params) {
             }
         })
         .then( function(ids) {
-            console.log("getAllCalendarIds then function running");
             return ids;
         })
         .catch( function() {
--- a/public/calendar_reboot.js
+++ b/public/calendar_reboot.js
@@ -203,16 +203,16 @@ CheckIt.prototype.signOut = function() {
 };
 
 // Triggers when there is a change in the storage.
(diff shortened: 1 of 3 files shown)


public/calendar_reboot.js
```diff
@@ -69,6 +72,19 @@ function CheckIt() {
...
+    // Displays the loading wheel.
+    var spinner = new Spinner().spin()
+     
+    // Hardcoded for now.
+    this.clearPage();
+    document.getElementById("loadingWheel").appendChild(spinner.el);
+};
+
+CheckIt.prototype.hideLoadingWheel = function() {
+    spinner.stop();
```
johntyree: This is an error.

`spinner` from `displayLoadingWheel` is not in scope here.

public/calendar_reboot.js
```diff
@@ -808,32 +826,63 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+    var event = params['event'];
     var allCalendarIdsKey = 'allCalendarIdsKey';
     //the current_active_calendar is the key for localStorage that stores
     //the active calendar's Id
     var current_active_calendar = 'current_active_calendar';
+    // This tells you whether the storage is actively working.
+    self.activeCalls = 0;
     
-    var toKey = function(id) {
+     var toKey = function(id) {
```
johntyree: ws

public/calendar_reboot.js
```diff
@@ -808,32 +826,63 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
         return key;
     };
     
+    self.startWork = function() {
+        // Will increment the counter and possibly fire an event.
+        console.log("Accessing storage");
+        self.activeCalls += 1;
+        
+        // Will dispatch the event backgroundActivityChange 
+        console.log("dipatching backgroundActivityChange");
```
johntyree: I think you only want to do this when `activeCalls` is nonzero. The event handler will have no way of knowing this the way it is currently written.

You *could* include the current value of `activeCalls` as data on the event object, but I don't think that's necessary for what you're trying to do.

public/calendar_reboot.js
```diff
@@ -808,32 +826,63 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
                 resolve(allCalendarIds);
             }
             else {
                 reject("Not found");
             }
         })
+        .then( function() {
+            
+            self.endWork();
+        }.bind(this))
```
johntyree: Binding to `this` won't break anything here, but realize that you're mixing up two different strategies for dealing with the problem. On the one hand, you bound `self` to the store object and are using it in all the methods by closing over it. On the other hand, you're binding `this` to (what will probably be) that same object. You can certainly do both, but it's a little confusing, as it implies to the reader (that's you as soon as you forget all these details) that `this` and `self` are probably different. Otherwise why go through the trouble of having both?

public/calendar_reboot.js
```diff
@@ -808,32 +826,63 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+        .catch( function() {
+            
+            self.endWork();
```
johntyree: The fact that you had to remember to duplicate this for both cases is precisely why Python has the `try ... except ... finally` syntax. https://docs.python.org/2.7/tutorial/errors.html#defining-clean-up-actions

public/calendar_reboot.js
johntyree: You have to manually pass along the arguments if you want them to appear in future calls to `.then()`.

    .then(function(ids) {
        self.endWork();
        return ids;
    })

public/calendar_reboot.js
johntyree: Don't forget to pass along the error message here as well.

public/index.html
```diff
@@ -144,6 +144,13 @@
 
 
 <div id="contentWrapper" class="container-fluid">
+
+
+    <div id="loadingWheel">
```
johntyree: Why not attach the event to this element instead of `document` or `window` ?  Not sure how that affects dispatching, it might not be worth it.

johntyree (changes requested):

public/calendar_reboot.js
```diff
@@ -35,9 +35,12 @@ function CheckIt() {
     
     var monthObjects;
     
+    this.backgroundActivityChange = new CustomEvent('backgroundActivityChange');
```
johntyree: I'm not sure what happens if you reuse a custom event, but I believe the typical way to do this is have the thing that will dispatch the event create a new one every time. This lets the dispatcher attach information to the event if necessary, for example, the number of active calls.

What you need to remember is the *type* of the event, which is just the string. That's all the listener needs to know in order to catch the event.

johntyree: And actually... after playing around with this for awhile and reading the docs, it occurs to me that no one else does it this way. Usually the dispatcher just provides a function can call if you want to hook a listener up. We already see this with `firebase.auth.onAuthStateChanged` for example. I'd do it that way.

Remember that you'll probably end up with more than one thing that needs to react to busy-state changes, so it's not enough to just remember a single callback. There will be many of them and you'll need to invoke them all.

I have some example code that does this, but you should give it a try first.

public/calendar_reboot.js
johntyree: I suggest also passing a string to `startWork` and `endWork` so that you can log where they were called from, otherwise it's difficult to track why things are happening the way are.

johntyree (changes requested):
I suggest merging your jitter branch into master and then merging master into this branch before continuing. The fact that everything is happening instantaneously is making it impossible to actually see the spin wheel because there's never any background activity.

You can always set the jitter to 0 if you want it to be fast again.

public/calendar_reboot.js
andrayantelo: What am I duplicating? And also what error message?

public/calendar_reboot.js
andrayantelo: What did you google for when you looked around to see what other people do?

public/calendar_reboot.js
andrayantelo: Do I make spinner a global variable? Or maybe a checkit variable?

public/calendar_reboot.js
andrayantelo: Hmm.. can you be more specific with "Usually the dispatcher just provides a function can call if you want to hook a listener up." How does it work? Or point me in the direction where I can read about it.

public/calendar_reboot.js
andrayantelo: Made it a checkit variable.

public/calendar_reboot.js
andrayantelo: Fixed.

public/calendar_reboot.js
johntyree: You had to duplicate the call to `endWork()` in two places (in `.then()` and in `.catch()`) because you don't know which of the two will happen.

The error message is the one that is passed to this function as an argument. Whatever you called `reject` with.

public/calendar_reboot.js
johntyree: The most recent and obvious example is `firebase.auth.onAuthStateChanged()`. You want to basically the same thing here. You have an object (`auth` or the `store`) that has some kind of hidden internal state that you don't really care about but you want to know when something important happens, such as logging in or out for `auth` or starting and finishing work for the `store`.

public/calendar_reboot.js
```diff
@@ -820,6 +872,27 @@ var LocalCalendarStorage = function(params) {
...
+        console.log("Accessing storage from " + processName);
+        self.activeCalls += 1;
+        
+        // Will dispatch the event backgroundActivityChange 
+        if (self.activeCalls > 0) {
+            console.log("dipatching backgroundActivityChange");
+        }
+    };
+    
+    self.endWork = function(accesor) {
```
johntyree: `s/accessor/processName`

public/calendar_reboot.js
```diff
@@ -820,6 +872,27 @@ var LocalCalendarStorage = function(params) {
...
     };
     
+
+    self.startWork = function(processName) {
+        // Will increment the counter and possibly fire an event.
+        console.log("Accessing storage from " + processName);
+        self.activeCalls += 1;
+        
+        // Will dispatch the event backgroundActivityChange 
+        if (self.activeCalls > 0) {
```
johntyree: Think about how the interaction between the store and the listeners will work. Is the store going to invoke the callbacks every time an operation starts? If so, you'll need to include the value of `activeCalls` somehow so that the listener can decide whether you're just starting to work or are doing more than one thing at a time.

What I mean is that the transition from 0 to 1 is semantically more important than say, going from 5 to 6, right? How will listeners know?

public/calendar_reboot.js
```diff
@@ -69,6 +71,52 @@ function CheckIt() {
 
 };
 
+CheckIt.prototype.setUpLoadingWheel = function() {
```
johntyree: Do you call this more than once? Why not just make it part of the constructor?

```javascript
this.spinner = new Spinner({
    lines: 13
  , length: 28
  ...
});
```

public/index.html
johntyree: I think you're going to discover that planting a giant spinner wheel in the middle of the UI everytime you touch storage is a bit too invasive.

public/calendar_reboot.js
```diff
@@ -69,6 +71,52 @@ function CheckIt() {
...
+    // is smaller.
+    this.spinner = new Spinner(opts);
+};
+
+CheckIt.prototype.displayLoadingWheel = function() {
+    // Displays the loading wheel.
+    var target = document.getElementById("loadingWheel");
+    // Hardcoded for now.
+    this.$buildFormAccordion.attr('hidden', true);
+    this.clearPage();
```
johntyree: So... every time I click a checkmark you're going to clear the page?

johntyree (changes requested):

public/calendar_reboot.js
andrayantelo: Yeah, I guess not. lol
I was only considering the time when a calendar is being loaded.

andrayantelo: I guess I won't need to be hiding the build form every time either.

public/calendar_reboot.js
andrayantelo: Yeah, I understand. You mean if I had a customEvent for example I'd add activeCalls to it's details right? So, that you can access it. Except I'm not using customEvents anymore. 

So, it would be ... in the function call... that you are listening to?

public/calendar_reboot.js
```diff
@@ -207,6 +208,8 @@ CheckIt.prototype.onActivityChanged = function(storageObj) {
     // Passing storageObj as argument to have access to activity calls, which
```
johntyree, line 197: Just pass the activity level directly if you want to do this. If you really want to access the entire store, you could have gotten it via `this.store` anyway.

johntyree (commented):
So far so good.

For bonus points, `store.onActivityChanged` should be callable multiple times to support having more than one callback function.

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
         //store the state in localStorage
 
         var stateP = new Promise(function(resolve, reject) {
+            //self.startWork();
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
...
             storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
             jitter(resolve);
         });
+        // When do I "endWork" here? in the stateP promise?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
...
         
         //put calendar in allCalendarIdss and store it
+        
         var idsP = self.getAllCalendarIds()
             .then(function (allCalendarIds) {
+                // start work here?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
...
                 allCalendarIds[calendarObj.state.uniqueId] = calendarObj.state.title;
                 storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
+                // end work here?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
...
             })
             .catch(function () {
                 console.log("No previous calendars in storage");
+                // start work?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -852,21 +971,28 @@ var LocalCalendarStorage = function(params) {
...
                 var allCalendarIds = {};
                 allCalendarIds[calendarObj.state.uniqueId] = calendarObj.state.title;
                 storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
+                // end work?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -886,6 +1012,8 @@ var LocalCalendarStorage = function(params) {
 
         return self.getAllCalendarIds()
             .then(function(allCalendarIds) {
+                // startwork?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -894,22 +1022,28 @@ var LocalCalendarStorage = function(params) {
                 removeFromLocalStorage(toKey(uniqueId));
                 // Remove active status from the calendar
                 removeFromLocalStorage(toKey(current_active_calendar));
+                
+                // end work?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -894,22 +1022,28 @@ var LocalCalendarStorage = function(params) {
...
             }.bind(this))
             .catch(function () {
                 console.log("Unable to remove calendar");
+                // end work?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -894,22 +1022,28 @@ var LocalCalendarStorage = function(params) {
...
             }.bind(this));
         
     };
     
     self.loadById = function(calendarObjId) {
         // Return a promise to a calendar object using its Id
         return new Promise( function(resolve, reject) {
+            //self.startWork(); // ?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -894,22 +1022,28 @@ var LocalCalendarStorage = function(params) {
...
             var calendar = loadFromLocalStorage(toKey(calendarObjId));
             
             if (calendar !== null) {
                 jitter(resolve, calendar);
+                //self.endWork(); // ?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -894,22 +1022,28 @@ var LocalCalendarStorage = function(params) {
...
             }
             else jitter(reject, "Calendar not found");
+            //self.endWork(); //?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -918,15 +1052,18 @@ var LocalCalendarStorage = function(params) {
         // Return a promise to the active_calendar id from storage
         
         return new Promise( function(resolve, reject) {
+            //self.startWork() //?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -918,15 +1052,18 @@ var LocalCalendarStorage = function(params) {
...
             var activeCalendarId = loadFromLocalStorage(toKey(current_active_calendar));
             
             if (activeCalendarId !== null ) {
                 
                 // Signal that the promise succeeded and make the value ready to go. 
                 jitter(resolve, activeCalendarId);
+                //self.endWork() //?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -918,15 +1052,18 @@ var LocalCalendarStorage = function(params) {
...
             }
             else {
                 jitter(reject, "Not found");
+                //self.endwork() //?
```
andrayantelo: ?

public/calendar_reboot.js
```diff
@@ -828,21 +933,35 @@ var LocalCalendarStorage = function(params) {
...
                 jitter(resolve, allCalendarIds);
+
             }
             else {
                 jitter(reject, "Not found");
             }
         })
+        .then( function(ids) {
+            console.log("getAllCalendarIds then function running");
+            self.endWork(); // should I actually be ending work here?
```
johntyree: Yes. The purpose of these is to indicate whether or not work is *still happening* and will finish sometime in the future. If you call `startWork` without calling `endWork` you will appear to be stuck in the "activity" state forever. You need to make sure that no matter what happens, `startWork` and `endWork` get called exactly the same number of times.

Also, don't forget to remove this comment.

public/calendar_reboot.js
johntyree: I explained the rational behind when you should or should not call `endWork` below.

public/calendar_reboot.js
```diff
@@ -69,6 +97,24 @@ function CheckIt() {
...
+
+CheckIt.prototype.displayLoadingWheel = function(elementId) {
+    // Displays the loading wheel.
+    
+    // Parameters: elementId 
+           // A string pertaining to the id of the element
+           // where you want to place the loading wheel.
+    
+    // Needs a plugin to use jQuery.
+    var target = document.getElementById(elementId);
```
johntyree: You can still user jQuery here and just grab the element out of it before you pass it to the spinner.
I think it's `this.spinner.spin($(selector).get());`.

public/calendar_reboot.js
```diff
@@ -820,12 +887,48 @@ var LocalCalendarStorage = function(params) {
...
+        self.activityChangeFunctions.forEach(function(func) {
+            func(self.activeCalls);
+        });
+        
+    };
+    
+    self.endWork = function() {
+        // Will decrement the counter and maybe fire an event.
+        
+        self.activeCalls -= 1;
```
johntyree, line 912: Can you add an assertion that `activeCalls >= 0` here? I have a feeling it will be tricky to make sure there is exactly one invocation of `endWork` for every `startWork`.

johntyree: Oh I see the log message now. I'd argue it should be `console.error` if `activeCalls < 0`.

public/index.html
johntyree: You have two divs with id `loadingWheel` ?

johntyree (changes requested):

public/calendar_reboot.js
```diff
@@ -104,9 +104,8 @@ CheckIt.prototype.displayLoadingWheel = function(elementId) {
     // Parameters: elementId 
            // A string pertaining to the id of the element
            // where you want to place the loading wheel.
+    var target = $('#loadingWheel').get()[0];
```
johntyree, line 107: `elementId` not `#loadingWheel`

public/calendar_reboot.js
```diff
@@ -899,7 +897,7 @@ var LocalCalendarStorage = function(params) {
         // Will increment the counter and possibly fire an event.
         
         self.activeCalls += 1;
-        console.log("store's startWork is running, activeCalls = " + self.activeCalls);
+        self.activeCalls += 1;
```
johntyree: What's going on here? You increment twice?

johntyree (changes requested):

public/calendar_reboot.js
```diff
@@ -911,8 +909,9 @@ var LocalCalendarStorage = function(params) {
     self.endWork = function() {
         // Will decrement the counter and maybe fire an event.
         
+        
         self.activeCalls -= 1;
-        console.log("store's endWork is running, activeCalls = " + self.activeCalls);
+        if (self.activeCalls < 0) console.error("No work has been started");
```
johntyree: It is best not to get into the habit of leaving out the curly braces. The benefit is very small and the likelihood of introducing bugs accidentally is very high. https://www.reddit.com/r/programming/comments/22ghj1/the_heartbleed_bug/

johntyree (changes requested):

johntyree (approved):

Issue #107, opened 2017-01-23
Securing user data


andrayantelo:
Users should only be able to read and write their own data.

PR #113, opened 2017-01-29
[WIP] Added new database rules

### LINKED ISSUE #111: Firebase storage security is too simple
Users can see all the calendars. Only security on it right now is that they have to be logged in.  They don't know whose calendar belongs to who. Fixes #107

andrayantelo:
Fixes #111 and #107


johntyree:
Getting to the point where we need to have some tests to make sure this all still works.

PR #114, opened 2017-01-29
Implementing the firebase storage and getting rid of the prefix "checkit_"

### LINKED ISSUE #110: Change over to firebase storage/database from local storage

andrayantelo:
Made a firebase storage object. Resolves #110 and #112. It gets initialized in the app method initFirebase and get passed auth. Has a few of the same methods as the localStorage store object.  Started the getAllCalendarIds method in the firebase storage object.  The firebase database once method doesn't do startWork and endWork like my methods do in the localStorage store object so I'm not sure how to use startWork and endWork in the getAllCalendarIds method yet.

Still WIP because I want to merge the new database rules into it, which means I will have to restructure how saves are made.

### DIFF
--- a/database.rules.json
+++ b/database.rules.json
@@ -9,16 +9,16 @@
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth !== null && data.parent().child('read/' + auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
+          ".read" : "auth !== null && data.parent().child('readers/' + auth.uid).exists()",
+          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()"
         },
-        "write" : {
+        "writers" : {
           ".read" : "auth !== null && data.child(auth.uid).exists()",
           ".write" : "auth !== null && data.child(auth.uid).exists()"
         },
-        "read" : {
+        "readers" : {
           ".read" : "auth !== null && data.child(auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
+          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()"
         }
       }
     }
--- a/database.rules.json
+++ b/database.rules.json
@@ -2,14 +2,16 @@
   "rules": {
     "users": {
       "$uid": {
-        ".read": "auth != null && auth.uid == $uid",
-        ".write": "auth != null && auth.uid == $uid"
+        ".read" : "auth!== null && auth.uid === $uid",
+        ".write" : "auth !== null && auth.uid === $uid"
       }
     },
-    "calendars": {
-      "$calendar_id": {
-      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
-      ".write": "auth !=null && data.child('permissionWrite/' + auth.uid.val()).exists()"
+    "calendars":{
+      "$calendarId": {
+        "calendarState": {
+          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
+          ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
+        } 
       }
     }
   } 
--- a/database.rules.json
+++ b/database.rules.json
@@ -9,7 +9,7 @@
     "calendars": {
       "$calendar_id": {
       ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
-      ".write": "auth !=null && data.child('permissionWrite').hasChild(root.child('users/'+auth.uid).val())"
+      ".write": "auth !=null && data.child('permissionWrite/' + auth.uid.val()).exists()"
       }
     }
   } 
--- a/database.rules.json
+++ b/database.rules.json
@@ -8,8 +8,8 @@
     },
     "calendars": {
       "$calendar_id": {
-      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid+'))",
-      ".write": "auth !=null && data.child('permissionWrite').hasChild(root.child('users/'+auth.uid+'))"
+      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
+      ".write": "auth !=null && data.child('permissionWrite').hasChild(root.child('users/'+auth.uid).val())"
       }
     }
   } 
--- a/database.rules.json
+++ b/database.rules.json
@@ -9,8 +9,8 @@
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
-          ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
+          ".read" : "auth!== null && data.parent().child('read/' + auth.uid).exists()",
+          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
         },
         "write" : {
           ".read" :,
--- a/database.rules.json
+++ b/database.rules.json
@@ -1,6 +1,16 @@
 {
   "rules": {
-    ".read": "auth != null",
-    ".write": "auth != null"
-  }
+    "users": {
+      "$uid": {
+        ".read": "auth != null && auth.uid == $uid",
+        ".write": "auth != null && auth.uid == $uid"
+      }
+    },
+    "calendars": {
+      "$calendar_id": {
+      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid+'))",
+      ".write": "auth !=null && data.child('permissionWrite').hasChild(root.child('users/'+auth.uid+'))"
+      }
+    }
+  } 
 }
--- a/database.rules.json
+++ b/database.rules.json
@@ -11,7 +11,15 @@
         "calendarState": {
           ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
           ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
-        } 
+        },
+        "write" : {
+          ".read" :,
+          ".write" :
+        }
+        "read" : {
+          ".read" :,
+          ".write" :
+        }
       }
     }
   } 
--- a/database.rules.json
+++ b/database.rules.json
@@ -2,23 +2,23 @@
   "rules": {
     "users": {
       "$uid": {
-        ".read" : "auth!== null && auth.uid === $uid",
+        ".read" : "auth !== null && auth.uid === $uid",
         ".write" : "auth !== null && auth.uid === $uid"
       }
     },
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth!== null && data.parent().child('read/' + auth.uid).exists()",
+          ".read" : "auth !== null && data.parent().child('read/' + auth.uid).exists()",
           ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
         },
         "write" : {
-          ".read" :,
-          ".write" :
-        }
+          ".read" : "auth !== null && data.child(auth.uid).exists()",
+          ".write" : "auth !== null && data.child(auth.uid).exists()"
+        },
         "read" : {
-          ".read" :,
-          ".write" :
+          ".read" : "auth !== null && data.child(auth.uid).exists()",
+          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
         }
       }
     }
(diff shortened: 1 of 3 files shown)


public/calendar_reboot.js
```diff
@@ -865,6 +870,95 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+            func(self.activeCalls);
+        });
+      
+    };
+    
+    self.getAllCalendarIds = function() {
+        // Returns a promise for the allCalendarIds object from storage
+        
+        var userId = self.user.currentUser.uid;
+        self.database.ref('/users/' + userId + '/checkit_allCalendarIds')
```
johntyree: This is the promise that you return. Call `startWork` before it starts and call `endWork` in the handlers instead of just logging.

johntyree (changes requested):

public/calendar_reboot.js
```diff
@@ -923,17 +923,58 @@ var firebaseCalendarStorage = function(user) {
...
+                return err;
+            })
         })
     };
+    
+    self.save = function(calendarObj) {
+        //save an App object (like a calendar object for example) in storage
+        
+        // Store the calendar state in the firebase database.
+
```
andrayantelo: @johntyree Do you think I should just have an objects in the /users/uid/ path like users/uid/calendars
and in there have an object like this:
{ calendar_uniqueId: { state object}}

Instead of the way it is now?

andrayantelo: Maybe users/uid/calUniqueId/calStateObject

johntyree: I think it will make things like sharing harder later if every calendar is "part of" a user, but if you're having trouble getting it to work at all like it is now, or you have a better reason to do it, then go for it.

public/calendar_reboot.js
```diff
@@ -865,6 +870,116 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+        // Store the user's allCalendarIds.
+        var updates = {};
+        updates['users/' + userId + '/checkit_allCalendarIds/' + calUniqueId] = calTitle;
+        
+        self.startWork();
+        return new Promise(function(resolve, reject) {
+        
+            self.getAllCalendarIds()
+                .then(function (allCalendarIds) {
+                    console.log("setting new value in database");
```
johntyree: log messages like this are a lot more useful if you include the relevant data

public/calendar_reboot.js
```diff
@@ -865,6 +870,116 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+                    firebase.database().ref().update(updates);
+                    console.log("DATA SET");
+                    self.endWork();
+                    resolve();  // At what point do I reject?
```
johntyree: > ` // At what point do I reject?`

You reject when there's an error. In this case, the only error that might happen will come from the firebase library. What happens when there's a firebase error? Which code gets executed?

johntyree (changes requested):

public/calendar_reboot.js
andrayantelo: the code inside of the catch function right?

andrayantelo: Pls respond @johntyree :cat:

andrayantelo: I don't need a reject then in this case?

public/calendar_reboot.js
```diff
@@ -136,20 +130,20 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
                        var calendar = new Calendar(state, this);
                        this.displayCalendar(calendar);
                    }
                }.bind(this))
-               .catch(function() {
-                   console.error("Calendar does not exist");
-                   this.store.removeActive();
+               .catch(function(err) {
+                   console.error("Calendar does not exist " + err);
+                   this.firebaseStore.removeActive();  // Should I be doing this here?
```
johntyree: It depends on the error probably. If the error is that the calendar doesn't exist then yes. If the error is because firebase couldn't load the data or the network was crappy and lost your request or something, then no.

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+        var calTitle = calendarObj.state.title;
+        var calState = calendarObj.state;
         
-        //put calendar in allCalendarIdss and store it
+        // Store the user's allCalendarIds and their calendarState .. setting the
+        // active calendar has it's own method.
+        var updates = {};
+        updates['users/' + userId + '/allCalendarIds/' + calUniqueId] = calTitle;
+        updates['calendars/' + calUniqueId + '/calendarState/'] = calState;
+        updates['calendars/' + calUniqueId + '/permissionRead/'] = userId;
```
johntyree: Remember to update this based on #116

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
-                allCalendarIds[calendarObj.state.uniqueId] = calendarObj.state.title;
-                storeInLocalStorage(toKey(allCalendarIdsKey), allCalendarIds);
-            })
-            
-        return Promise.all([stateP, idsP]);
+        self.startWork();
+        return new Promise(function(resolve, reject) {
+            self.database.ref().update(updates);
+            self.endWork();
+            resolve();  // Still unsure whether I will need to use reject in here or not.
```
johntyree: The call to `update()` above is asynchronous...

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+    self.setActiveById = function(calendarObjId) {
+        // Set the active calendar/object by using its Id
+        
+        var userId = self.user.uid;
+        var updates = {};
+        updates['users/' + userId + '/currentActiveCalendar/'] = calendarObjId;
+        
+        self.startWork();
+        return new Promise( function(resolve, reject) {
+            self.database.ref().update(updates);
```
johntyree: Again this is an asynchronous call, so this whole function isn't going to behave the way you want.
https://firebase.google.com/docs/reference/js/firebase.database.Reference#update

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+            self.endWork();
+            resolve();  // Reject somewhere?
```
johntyree: > `// Reject somewhere?`

You want to `reject()` iff there's an error that the caller could `.catch()` and do something with. What might be an error here? Where would it come from?

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
-                // Remove the calendar from local storage
-                removeFromLocalStorage(toKey(uniqueId));
-                // Remove active status from the calendar
-                removeFromLocalStorage(toKey(current_active_calendar));
-                
-            })
-            .catch(function () {
-                console.log("Unable to remove calendar");
-            });
+        // Remove a calendar from storage by using it's Id.
```
johntyree: its

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+        // First deleting it from /user/uid/allCalendarIds
+        // Second remove permissions
+        // third removing it from calendars/
+        // removing currentActive status has it's own method
```
johntyree: its

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+        var userId = self.user.uid;
         
+        self.startWork();
+        return new Promise(function(resolve, reject) {
```
johntyree: This is all broken because the `.remove()` calls are asynchronous. What is going to be the order of events in this function?

public/calendar_reboot.js
```diff
@@ -921,151 +933,168 @@ var LocalCalendarStorage = function(params) {
...
+        
+        self.startWork();
         return new Promise( function(resolve, reject) {
-            var calendar = loadFromLocalStorage(toKey(calendarObjId));
-            
-            if (calendar !== null) {
-                jitter(resolve, calendar);
-            }
-            else jitter(reject, "Calendar not found");
+            self.database.ref('users/' + userId + '/currentActiveCalendar/').remove();
```
johntyree: Async...

johntyree (changes requested):
After #116 lands, merge master back into this branch to get the updated rules and database files. Then make sure all your paths still make sense.

database.rules.json
```diff
@@ -1,6 +1,26 @@
...
+        ".write" : "auth !== null && auth.uid === $uid"
+      }
+    },
+    "calendars":{
+      "$calendarId": {
+        "calendarState": {
+          ".read" : "auth !== null && data.parent().child('read/' + auth.uid).exists()",
+          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
+        },
+        "write" : {
```
johntyree: I feel like "write" and "read" here could have better names. Maybe `writers` and `readers` or `can_write` and `can_read`.

public/calendar_reboot.js
```diff
@@ -136,20 +130,20 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
                        var state = activeCalendarState;
                        var calendar = new Calendar(state, this);
                        this.displayCalendar(calendar);
                    }
                }.bind(this))
-               .catch(function() {
-                   console.error("Calendar does not exist");
-                   this.store.removeActive();
+               .catch(function(err) {
+                   console.error("Calendar does not exist " + err);
```
johntyree: Are you certain this means that the calendar doesn't exist? What if firebase itself just had a temporary problem?

public/calendar_reboot.js
```diff
@@ -865,24 +874,29 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+    
+    var jitter = function(func, arg) {
+        var runFunc = function () {
+            func(arg);
+            self.endWork();
+        };
         
-        var key = prefix + "_" + id;
-        return key;
+        var randomNumber = Math.random() * 4000;
```
johntyree: You probably don't want a 4s delay in master.

public/calendar_reboot.js
```diff
@@ -921,151 +933,187 @@ var LocalCalendarStorage = function(params) {
...
         //save an App object (like a calendar object for example) in storage
         
-        //store the state in localStorage
-
-        var stateP = new Promise(function(resolve, reject) {
-            storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
-            jitter(resolve);
-        });
+        // Store the calendar state in the firebase database.
+        // TO DO save the actual calendar state.
```
johntyree: TODO not "TO DO"

Also isn't this being done?

public/calendar_reboot.js
```diff
@@ -921,151 +933,187 @@ var LocalCalendarStorage = function(params) {
...
         
+        var userId = self.user.uid;
+        var calUniqueId = calendarObj.state.uniqueId;
+        var calTitle = calendarObj.state.title;
+        var calState = calendarObj.state;
         
-        //put calendar in allCalendarIdss and store it
+        // Store the user's allCalendarIds and their calendarState .. setting the
+        // active calendar has it's own method.
```
johntyree: its

Also I don't understand these two comments. Did you accidentally

public/calendar_reboot.js
```diff
@@ -921,151 +933,187 @@ var LocalCalendarStorage = function(params) {
...
+        .catch(function(err) {
+            console.error("Unable to set active Id: " + err);
+            return err;
+        })
     };
     
     self.remove = function(calendarObj) {
-        //remove an app object (like a calendar) from storage
+        // Remove an app object (like a calendar) from storage
+        // To do: Determine if this method is necessary
```
johntyree: TODO

public/calendar_reboot.js
```diff
@@ -921,151 +933,187 @@ var LocalCalendarStorage = function(params) {
...
+        
         var uniqueId = calendarObj.state.uniqueId;
         self.removeById(uniqueId);
     };
     
     self.removeById = function(uniqueId) {
-        //remove a calendar from storage by using it's Id.
+        // Remove a calendar from storage by using its Id.
+        // First deleting it from /user/uid/allCalendarIds
+        // Second remove permissions
```
johntyree: This doesn't seem accurate.

johntyree (changes requested):

public/calendar_reboot.js
andrayantelo: Do I even need jitter anymore?

public/calendar_reboot.js
andrayantelo: Forgot to delete that comment.

public/calendar_reboot.js
andrayantelo: You don't have to save the currentActiveCalendar in the save method because setActive is its own method. I was just writing that there so that I could remember that.

public/calendar_reboot.js
```diff
@@ -136,20 +130,20 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
                        var state = activeCalendarState;
                        var calendar = new Calendar(state, this);
                        this.displayCalendar(calendar);
                    }
                }.bind(this))
-               .catch(function() {
-                   console.error("Calendar does not exist");
-                   this.store.removeActive();
+               .catch(function(err) {
+                   console.error("Could not load calendar " + err);
```
johntyree, line 145: extra space between "not" and "load" :smiling_imp:

public/calendar_reboot.js
```diff
@@ -136,20 +130,20 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
+                   this.firebaseStore.removeActive();
```
johntyree: trailing whitespace

johntyree (changes requested):

public/calendar_reboot.js
andrayantelo: No there isn't :angel:

johntyree (approved):

PR #116, opened 2017-01-31
[WIP] Update sample_calendar.json and database.rules.json

andrayantelo:
Is this a good way to structure it to manage authorization for reading and writing to different calendars?
Addresses issue #107

### DIFF
--- a/database.rules.json
+++ b/database.rules.json
@@ -2,14 +2,16 @@
   "rules": {
     "users": {
       "$uid": {
-        ".read": "auth != null && auth.uid == $uid",
-        ".write": "auth != null && auth.uid == $uid"
+        ".read" : "auth!== null && auth.uid === $uid",
+        ".write" : "auth !== null && auth.uid === $uid"
       }
     },
-    "calendars": {
-      "$calendar_id": {
-      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
-      ".write": "auth !=null && data.child('permissionWrite/' + auth.uid.val()).exists()"
+    "calendars":{
+      "$calendarId": {
+        "calendarState": {
+          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
+          ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
+        } 
       }
     }
   } 
--- a/database.rules.json
+++ b/database.rules.json
@@ -9,7 +9,7 @@
     "calendars": {
       "$calendar_id": {
       ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
-      ".write": "auth !=null && data.child('permissionWrite').hasChild(root.child('users/'+auth.uid).val())"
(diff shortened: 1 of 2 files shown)


sample_calendar.json
```diff
@@ -1,25 +1,38 @@
...
+        "checkit_148460826559529600" { :
+            {
             "startDateString":"20170101",
             "endDateString":"20170202",
             "title":"hi",
             "uniqueId":148460826559529600,
             "checkedDays":{}
             },
+            "permissionRead": "JULxYGaHuFYlSkXqTC2Fx9KIofp2",
+            "permissionWrite": "JULxYGaHuFYlSkXqTC2Fx9KIofp2"
```
andrayantelo: I could also do:
"permissionRead" : {
    "JULxYGaHuFYlSkXqTC2Fx9KIofp2" : "true"
}

Although, I'm not sure if that would make sense? I mean, will it ever be false?

database.rules.json
```diff
@@ -2,14 +2,16 @@
...
       }
     },
-    "calendars": {
-      "$calendar_id": {
-      ".read": "auth !=null && data.child('permissionRead').hasChild(root.child('users/'+auth.uid).val())",
-      ".write": "auth !=null && data.child('permissionWrite/' + auth.uid.val()).exists()"
+    "calendars":{
+      "$calendarId": {
+        "calendarState": {
+          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
```
johntyree: Two things here.

First, you shouldn't ever need to do `=== true`. Think about why.

Second, I think the value of the key doesn't really matter anyway. You're not going to put in uids with values that are anything other than `true` right? What about `".read": "auth !== null && data.parent().child('read/' + auth.uid).exists()"` ?

johntyree (changes requested):

database.rules.json
```diff
@@ -9,9 +9,17 @@
...
-          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
-          ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
-        } 
+          ".read" : "auth!== null && data.parent().child('read/' + auth.uid).exists()",
+          ".write" : "auth !== null && data.parent().child('write/' + auth.uid).exists()"
+        },
+        "write" : {
+          ".read" :,
+          ".write" :
+        }
```
johntyree: Missing a comma

database.rules.json
```diff
@@ -9,9 +9,17 @@
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth!== null && data.parent().child('read').hasChild(auth.uid) && data.parent().child('read').child(auth.uid).val() === true",
-          ".write" : "auth !== null && data.parent().child('write').hasChild(auth.uid) && data.parent().child('write').child(auth.uid).val() === true"
-        } 
+          ".read" : "auth!== null && data.parent().child('read/' + auth.uid).exists()",
```
johntyree: Missing a space after first `auth`

johntyree (changes requested):
https://jsonformatter.curiousconcept.com/

johntyree (approved):

PR #125, opened 2017-02-06
Changed name of public directory to docs and moved CNAME in there

andrayantelo:
Fixes #123


andrayantelo:
Appears to have worked @johntyree

Issue #115, opened 2017-01-30
Write tests 


andrayantelo:
Raise exceptions for emptyCalendarState and generateMonthObjects?

johntyree:
There's at least two levels to this. We want to test that the app behaves correctly *assuming* that the store is also working correctly. For that, it will be easiest to use local storage.

We also want to test that we've set up firebase correctly and are using it correctly.
Some ideas:
- [x] It shouldn't be possible to read or write to a calendar for which you don't have permission
- [x] Saving a calendar and loading it again should give you back exactly what you started with
- [ ] Deleting all users should leave the database totally empty
- [x] Deleting a calendar should remove it from the user
- [ ] The calendar itself should also be deleted if there are no remaining users with permission to read/write it
- [ ] There should always be at least one user with write permission on a calendar
- [x] All users with write permission should also have read permission (maybe it's best to just treat write permission as readwrite?)

andrayantelo:
All of the firebase stuff is just stuff you can test on the firebase console, right? 

As for testing the app (while using localStorage), what kinds of things would I be testing for? Like if when you make a calendar object the right things are defined and stuff like that? So, I guess I want to do unit testing, so test each function and make sure that it does what it's supposed to do. Should I use unit.js? mocha? karma (is that just for angular js?) @johntyree

johntyree:
Use whichever testing framework you are comfortable with. jQuery probably has something for this if you don't have any preference.

Some of the things I listed above can tested against local storage as well. It's valuable to make sure the local storage backend still works in case we end up wanting to use some hybrid of the two, for example. Also, if you want to test more complicated features locally, you must first be sure that all the simple things still work.

Note that you *should* now be keeping track of `/calendars/$calid/{readers,writers}` for *local storage* too and can check that they are written to and read from correctly when doing things.

PR #128, opened 2017-02-07
[WIP] Writing tests for checkit 

andrayantelo:
So far have only written a few tests using QUnit. Addresses #115

### DIFF
--- a/docs/store.js
+++ b/docs/store.js
@@ -306,10 +306,12 @@ var LocalCalendarStorage = function(params) {
     self.getAllCalendarIds = function() {
         // Returns a promise for the allCalendarIds object from storage
         
+        self.startWork();
         return new Promise( function(resolve, reject) {
-            var allCalendarIds = loadFromLocalStorage(toKey(allCalendarIdsKey));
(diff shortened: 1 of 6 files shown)


docs/store.js
```diff
@@ -329,23 +330,22 @@ var LocalCalendarStorage = function(params) {
         
         //store the state in localStorage
         var stateP = new Promise(function(resolve, reject) {
-            storeInLocalStorage(toKey(calendarObj.state.uniqueId), calendarObj.state);
+            localStorage.setItem(toKey(calendarObj.state.uniqueId), JSON.stringify(calendarObj.state));
```
johntyree, line 333: Since you have to remember to do a few different things for this to work correctly, (`toKey`... `stringify`...) making a little internal helper function would have been better here. Then you don't have as much copy paste noise all over the place, but the local storage is still kept internal to the store.

```
self.set = function(key, val) {
   localstorage.setItem(toKey(key), JSON.stringify(val));
}
```

johntyree:
This PR is doing too many things, in my opinion. You're reintroducing local storage. Trying to set up QUnit. And refactoring how local storage is used all at once.

PR #130, opened 2017-02-07
[WIP] Switch to using local Storage

andrayantelo:
Get error about firebase database permissions even though I switched to using localStorage store manager.


johntyree:
I'm confused about the purpose of this. It still depends on firebase working, but just uses local storage to actually store the data?

PR #131, opened 2017-02-07
Refactoring the code for the local Storage manager

andrayantelo:
Getting rid of the localStorage helper functions in checkit.js. Moving a few to localStorage manager object.

### DIFF
--- a/docs/store.js
+++ b/docs/store.js
@@ -260,12 +260,16 @@ var LocalCalendarStorage = function(params) {
         localstorage.setItem(toKey(key), JSON.stringify(val));
     };
     
-    self.getFromLocalStorage = function(key) {
+    self.getFromStorage = function(key) {
         var storageItem = localStorage.getItem(toKey(key))
         if (storageItem !== null) {
             storageItem = JSON.parse(storagItem);
         }
         return storageItem;
+    };
+    
+    self.removeFromStorage = function(key) {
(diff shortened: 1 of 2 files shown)


docs/store.js
```diff
@@ -256,6 +256,23 @@ var LocalCalendarStorage = function(params) {
...
     };
     
+    self.setInStorage = function(key, val) {
+        localstorage.setItem(toKey(key), JSON.stringify(val));
+    };
+    
+    self.getFromStorage = function(key) {
+        var storageItem = localStorage.getItem(toKey(key))
+        if (storageItem !== null) {
+            storageItem = JSON.parse(storagItem);
```
johntyree: Typo

docs/store.js
```diff
@@ -256,6 +256,23 @@ var LocalCalendarStorage = function(params) {
         return key;
     };
     
+    self.setInStorage = function(key, val) {
```
johntyree: These functions shouldn't be on self. No one outside storage should use them

johntyree (changes requested):

docs/store.js
andrayantelo: Fixed

docs/store.js
andrayantelo: Fixed

johntyree (approved):

Issue #117, opened 2017-01-31
Write firebase database validation rules for data


andrayantelo:
Certain data needs certain structure. For example, calendar titles need to be strings, dates need to be dates ( currently they need to be strings but may need to change this ).

johntyree:
I think you mean validation?

https://firebase.google.com/docs/database/security/#section-validation

andrayantelo:
Yes, I mean validation.

andrayantelo:
Do I want to save the user's email in the users/user.uid/ section of the database?
What does it mean to define database indices?

`"calendarState": {
    ".read" : "auth !== null",
     ".write" : "auth !== null && (!data.exists() || root.child('calendars/' +$calendarId + '/writers/' + auth.uid).exists())"
       },`

Why did `!newData.exists()` not work?

PR #133, opened 2017-02-07
Validate data

andrayantelo:
Working on validation rules as well as fixing read and write rules. Addresses #117

### DIFF
--- a/database.rules.json
+++ b/database.rules.json
@@ -9,27 +9,14 @@
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth !== null && data.parent().child('readers/' + auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()",
-          "startDateString": {
-            ".validate" : "newData.isString() && newData.val().length > 0"
-          },
-          "endDateString": {
-            ".validate" : "newData.isString() && newData.val().length > 0"
-          },
-          "title":{
-            ".validate": "newData.isString() && newData.val().length > 0"
-          }
-        },
-        "readers" : {
-          ".read" : "auth !== null && data.child(auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()"
-        },
+          ".read" : "auth !== null",
+          ".write" : "auth !== null && (!data.exists() || root.child('calendars/' +$calendarId + '/writers/' + auth.uid).exists())"
+       },
         "writers" : {
-          ".read" : "auth !== null && data.child(auth.uid).exists()",
-          ".write" : "auth !== null && root.child('users/' + auth.uid + '/allCalendarIds/' + $calendarId).exists()"
+          ".write" : "auth !== null && !data.exists()",
+          ".read": "auth !== null"
         }
       }
     }
-  } 
+  }
 }
--- a/database.rules.json
+++ b/database.rules.json
@@ -10,7 +10,16 @@
       "$calendarId": {
         "calendarState": {
           ".read" : "auth !== null && data.parent().child('readers/' + auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()"
+          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()",
+          "endDateString" : {
+            ".validate": "newData.isString()"
+          },
+          "startDateString" : {
+            ".validate": "newData.isString()"
+          },
+          "title" : {
+            ".validate" : "newData.isString()"
+          }
         },
         "writers" : {
           ".read" : "auth !== null && data.child(auth.uid).exists()",
--- a/database.rules.json
+++ b/database.rules.json
@@ -2,10 +2,10 @@
   "rules": {
     "users": {
       "$uid": {
+        ".read" : "auth !== null && auth.uid === $uid",
         "allCalendarIds" : {
           "$calendarId" : {
-            ".read" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/readers/' + auth.uid).exists())",
-            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists())",
+            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + $uid).exists())",
             ".validate" : "newData.isString() && newData.val().length > 0 ",
               "$calendarTitle" : {
                 ".validate" : "newData.isString() && newData.val().length > 0"
@@ -13,11 +13,8 @@
           }
         },
         "currentActiveCalendar" : {
-          "$calendarId" : {
-            ".read" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/readers/' + auth.uid).exists())",
-            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists())",  
+            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + newData.val() + '/writers/' + auth.uid).exists())",  
             ".validate" : "newData.isString() && newData.val().length > 0"
-          }
         }
       }
     },
--- a/database.rules.json
+++ b/database.rules.json
@@ -48,14 +48,14 @@
           ".write" : "auth !== null && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists())",
           ".read": "auth !== null && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists()) ",
           "$uid" : {
-            ".validate": "newData.isString() && newData.val().length > 0"
+            ".validate": "newData.val() === true || newData.val() === false"
           }
         },
         "readers" : {
           ".write" : "auth !== null && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists())",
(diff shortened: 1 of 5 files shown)


docs/store.js
```diff
@@ -91,13 +91,28 @@ var firebaseCalendarStorage = function(params) {
         var calTitle = calendarObj.state.title;
         var calState = calendarObj.state;
         
-        // Store the user's allCalendarIds and their calendarState .. setting the
-        // active calendar has its own method.
+        // Store the calendar into allCalendarIds first so that the user
+        // has permission to write into calendars/calUniqueId
```
johntyree: Maybe you should be allowed to create a new calendar as long as the final state is that the calendar is also in your users list of cal ids?

docs/store.js
```diff
@@ -91,10 +91,13 @@ var firebaseCalendarStorage = function(params) {
...
         var calState = calendarObj.state;
         
-        // Store the calendar into allCalendarIds first so that the user
-        // has permission to write into calendars/calUniqueId
+        // Store the calendar into allCalendarIds, store calState
+        // TODO work on save method, how to save readers, writers, permissions, 
+        // states, etc.
         var updates = {};
         updates['users/' + userId + '/allCalendarIds/' + calUniqueId] = calTitle;
+        updates['calendars/' + calUniqueId + '/calendarState'] = calState;
```
johntyree: Not important, but you can also just define all these when you define the object literal. Not necessarily better or worse.

```
var updates = {
  'users/' + userId + '/allCalendarIds/' + calUniqueId: calTitle,
 ....
}
```

andrayantelo: Oh okay. Will keep that in mind.

database.rules.json
```diff
@@ -3,24 +3,57 @@
     "users": {
       "$uid": {
         ".read" : "auth !== null && auth.uid === $uid",
-        ".write" : "auth !== null && auth.uid === $uid"
+        "allCalendarIds" : {
+          //".write" : "true",
```
johntyree: Remove this

database.rules.json
```diff
@@ -3,24 +3,57 @@
...
+          "$calendarId" : {
+            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + $uid).exists())",
+            ".validate" : "newData.isString() && newData.val().length > 0"
+          }
+        },
+        "currentActiveCalendar" : {
+            ".write" : "auth !== null && auth.uid === $uid && (!data.exists() || root.child('calendars/' + newData.val() + '/writers/' + auth.uid).exists())",
```
johntyree: Really you only need to be able to read the calendar to show it.

johntyree: Not important now, but if it's supposed to be possible to share a calendar somehow then we'll have to look at this again

database.rules.json
```diff
@@ -3,24 +3,57 @@
...
     "calendars":{
       "$calendarId": {
         "calendarState": {
-          ".read" : "auth !== null && data.parent().child('readers/' + auth.uid).exists()",
-          ".write" : "auth !== null && data.parent().child('writers/' + auth.uid).exists()"
-        },
+          ".read" : "auth !== null && (!data.exists() || root.child('calendars/' + $calendarId + '/readers/' + auth.uid).exists())",
+          ".write" : "auth !== null && (!data.exists() || root.child('calendars/' +$calendarId + '/writers/' + auth.uid).exists())",
+          "startDateString": {
+            ".validate" : "newData.isString() && newData.val().length > 0 && newData.val().matches(/\\d{8}/)"
```
johntyree: I think the regex alone is enough here. If it's not a string it will fail. If it is, then the length will be restricted to exactly 8.

database.rules.json
```diff
@@ -3,24 +3,57 @@
...
+          },
+          "endDateString": {
+            ".validate" : "newData.isString() && newData.val().length > 0 && newData.val().matches(/\\d{8}/)"
+          },
+          "title":{
+            ".validate": "newData.isString() && newData.val().length > 0 && newData.val().length < 256"
+          },
+          "uniqueId" : {
+            ".validate" : "newData.isString() && newData.val().length > 0"
```
johntyree: Probably something like `newData.val().matches(/([A-Za-z0-9]+/)` is better here. The id should be just be some blob of alphanumeric something or another.

database.rules.json
```diff
@@ -3,24 +3,57 @@
...
+          "checkedDays" : {
+            "$dateString" : {
+              ".validate" : "$dateString.matches(/\\d{8}/) && (newData.val() === 1 || newData.val() === 0)"
+            }
+          }
+       },
         "writers" : {
-          ".read" : "auth !== null && data.child(auth.uid).exists()",
-          ".write" : "auth !== null && data.child(auth.uid).exists()"
+          ".write" : "auth !== null && (!data.exists() || root.child('calendars/' + $calendarId + '/writers/' + auth.uid).exists())",
```
johntyree: You're already here.  `data.child(auth.uid).exists()` ?  Maybe `data.parent.child` at worst.

docs/store.js
```diff
@@ -81,6 +81,69 @@ var firebaseCalendarStorage = function(params) {
...
     
+    self.addWriter = function(user, calObj) {
+        // Add a user to the writers directory in the database.
+        
+        var updates = {};
+        updates['calendars/' + calObj.state.uniqueId + '/writers/' + user.uid] = true;
+        
+        self.startWork();
+        return self.database.ref().update(updates)
+        .then(function() {
```
johntyree: fix the indentation here please

docs/store.js
```diff
@@ -81,6 +81,69 @@ var firebaseCalendarStorage = function(params) {
...
+            return err
+        })
+        
+    };
+    
+    self.removeWriter = function(user, calObj) {
+        // Remove a user form the writers directory in the database
+        
+        self.database.ref('calendars/' + calObj.state.uniqueId + '/writers/' + user.uid).remove()
+        .then(function() {
```
johntyree: indentation on all of these.

docs/store.js
```diff
@@ -91,24 +154,24 @@ var firebaseCalendarStorage = function(params) {
         var calTitle = calendarObj.state.title;
         var calState = calendarObj.state;
         
-        // Store the user's allCalendarIds and their calendarState .. setting the
-        // active calendar has its own method.
+        // Store the calendar into allCalendarIds, store calState
+        // TODO work on save method, how to save readers, writers, permissions,
```
johntyree: Is this out of date now?

docs/store.js
```diff
@@ -230,6 +294,26 @@ var firebaseCalendarStorage = function(params) {
...
+        // as a writer and a reader, sets the calendar as the currentActiveCalendar
+        // and then runs the save method (store cal state and allCalendarIds
+        
+        self.startWork();
+        var addWriterP = self.addWriter(self.user, calendarObj)
+        .then(function() {
+            self.addReader(self.user, calendarObj);
+            self.setActiveById(calendarObj.state.uniqueId);
+            self.save(calendarObj);
+            self.endWork();
```
johntyree: You're not done with work until those other promises return.  You need a `Promise.all` here.

johntyree (changes requested):

database.rules.json
andrayantelo: Done.

database.rules.json
andrayantelo: Okay.

andrayantelo: Changed it to readers.

database.rules.json
andrayantelo: Left just the regex.

database.rules.json
andrayantelo: Used that as uniqueId's validation rule.

database.rules.json
andrayantelo: Changed it to the first option.

docs/store.js
andrayantelo: Fixed.

docs/store.js
andrayantelo: Fixed.

docs/store.js
andrayantelo: Yes. Removed.

docs/store.js
andrayantelo: Done.

docs/store.js
```diff
@@ -210,21 +207,21 @@ var firebaseCalendarStorage = function(params) {
...
+                    self.endWork();
+                })
+                .catch(function(err) {
+                    console.error('Unable to remove calendar: ' + err);
+                    self.endWork();
+                })
             })
             .catch(function(err) {
-                console.error('Unable to remove calendar: ' + err);
+                console.error('Unable to remove calendar from user\'s allCalendarIds: ' + err);
```
johntyree: It's usually easier to read if you make use of the other type of quotes so that you don't have to escape.

docs/store.js
```diff
@@ -302,16 +299,24 @@ var firebaseCalendarStorage = function(params) {
...
-        .catch(function(err) {
-            console.error("Problems initializing calendar " + err);
-            self.endWork();
-        });
+            .then(function() {
+                var addReaderP = self.addReader(self.user, calendarObj);
+                var setActiveP = self.setActiveById(calendarObj.state.uniqueId);
+                var saveP = self.save(calendarObj);
+                Promise.all([addReaderP, setActiveP, saveP])
+                .then(function() {
```
johntyree: I find that lining up the methods vertically if they are going to operate on the same object is an easy way to organize it.
```
Promise.all([...])
       .then(function() {
          ...
       })
       .catch(function() {});
```
However you decide to organize it, it should be such that lines that begin in the column indicate new *statements*, not expressions. If an expression continues onto a second line, it should be indented somehow to give a visual indication that this is all a single computation that will return a single value.

johntyree (changes requested):
This is probably good to go. I only had some cosmetic suggestions.

docs/store.js
andrayantelo: Do you mean back ticks?

PR #134, opened 2017-02-08
Writing tests for Month and Calendar Objects and their helper functions

andrayantelo:


johntyree:
What am I reviewing here?

andrayantelo:
I haven't gotten started on this one yet.

tests/tests.html
```diff
@@ -0,0 +1,67 @@
...
+  <!-- include font awesome -->
+  <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
+
+  <!-- Material Design -->
+  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+
+  <!-- Loading spinner -->
+  <script src="https://cdn.jsdelivr.net/spinjs/2.3.2/spin.min.js"></script>
+  
+  <script src="https://www.gstatic.com/firebasejs/3.6.5/firebase.js"></script>
```
johntyree: You don't want to be running tests against your production database.

Either use localstorage or figure out how to use firebase's testing facilities.

tests/tests.js
```diff
@@ -0,0 +1,259 @@
+// Tests for checkit app
+
+// the test method gets passed a string as the first argument (naming the
+// test), and a function as the second (which will run the code for this test)
+// div #qunit-fixture contains an extract of the markup from our app, enough to 
+// write useful tests against. By putting it in this div, we don't have to worry
+// about DOM changes from one test affecting other tests because qunit will 
+// automatically reset the markup after each test.
+
+QUnit.test( "a basic test example", function( assert ) {
```
johntyree: Probably don't need this one :)

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+    assert.equal(this.testmonth.firstDayIndex, 3, "Checking index of the first of the month");
+    
+    assert.deepEqual(this.testmonth.dayIndex, {}, "Initialized the day Index object");
+    assert.equal(this.testmonth.monthId, "20171", "Initialized the month Id");
+});
+
+
+// Testing Calendar Helper functions
+QUnit.module( "Calendar helper functions tests", {
+});
```
johntyree, line 44: Is this supposed to be around these tests?

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+
+QUnit.test("generateUniqueId test", function(assert) {
+    assert.expect(1);
+    var uniqueId = generateUniqueId();
+    var uniqueId2 = generateUniqueId();
+    
+    assert.notEqual(uniqueId, uniqueId2);
+    // Anything else I should be testing in here?
```
johntyree: Seems ok

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+    assert.equal(this.calendar.state.endDateString, "20170320", "Check calendar endDateString added a month");
+    assert.equal(this.calendar.monthObjects.length, 2, "Check month was added to monthObjects array");
+    assert.equal(typeof(this.calendar.addMonth()), "object", "addMonth returns an object");
+});
+
+QUnit.test("generateMonthObjects calendar method test", function(assert) {
+    
+    assert.expect(5);
+    
+    var monthObjects = this.calendar.generateMonthObjects(moment("20170101", "YYYYMMDD"), moment("20170214", "YYYYMMDD"));
```
johntyree: You also want to test corner cases and errors. What if the start and end dates are the same? What if they are one day apart? What if the end date is earlier than start date? Does the function return the "right" answer in all these cases?

johntyree (changes requested):

tests/tests.js
andrayantelo: No.

andrayantelo: It's weird.

tests/tests.html
andrayantelo: If I don't include this script, then I can't use auth either. Should I not use auth either?

tests/template.html
```diff
@@ -0,0 +1,166 @@
...
+        <td>
+          <div class="nill"></div>
+        </td>
+    </tr>
+    <tr class="week">
+      <td>
+        <div class="nill"></div>
+      </td>
+      <td>
+        <div class="nill"></div>
```
johntyree: DIdn't this get changed to `.nil` at some point?

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+
+QUnit.test("emptyCalendarState test", function(assert) {
+    assert.expect(6);
+    var params = {startDate: "2017-02-14" , endDate: "2017-02-20" , calendarTitle: ""};
+    var state = emptyCalendarState(params);
+    
+    assert.equal(state.title, "", "Checking state title");
+    assert.equal(state.startDateString, "20170214", "Checking startDateString");
+    assert.equal(state.endDateString, "20170220", "Checking endDateString");
+    // What kind of assertion for uniqueId?
```
johntyree: non-empty is fine

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+    assert.deepEqual(state.checkedDays, {});
+    
+    //assert that an exception is thrown with invalid dates.
+    var params = {startDate: "2017-02-14", endDate: "2017-01-01", calendarTitle: "hello"};
+    // I don't understand the qunit api for assert.raises (or assert.throws)
```
johntyree: What's your question?

tests/tests.js
```diff
@@ -0,0 +1,259 @@
...
+    assert.expect(7);
+    
+    var monthObjects = this.calendar.generateMonthObjects(moment("20170101", "YYYYMMDD"), moment("20170214", "YYYYMMDD"));
+
+    assert.ok(Array.isArray(monthObjects));
+    assert.equal(monthObjects[0].dateString, "20170101");
+    assert.equal(monthObjects[1].dateString, "20170201");
+    assert.equal(monthObjects[1].lastActiveDay, 14);
+    assert.equal(monthObjects.length, 2);
+    // Don't understand why this assertion passes
```
johntyree: These assertions check to make sure that your function fails as expected. The assertion succeeds because the function failed in the way you described.

johntyree (changes requested):
If you aren't explicitly testing the database you probably don't want to be initializing and connecting to firebase every time you run your tests.

You should be able to test your code against a mock database and test the database separately.

tests/tests.js
andrayantelo: How does it work? Although, I followed along with the example they had when I wrote this.

tests/template.html
andrayantelo: Fixed.

docs/checkit.js
```diff
@@ -732,7 +742,8 @@ CheckIt.prototype.findCurrentDay = function() {
 
 $(document).ready(function() {
     
-    checkit = new CheckIt();
+    checkit = new CheckIt('localStorage');
```
johntyree: Won't this break everything? I thought we only wanted to use localStorage in tests.

johntyree: Oh I see the problem, this is in `checkit.js`. You probably want this module to have all the code you need to *define* a Checkit, but not actually create one. The code to create the object can either be in a `main.js` or `app.js` or whatever the cool kids are calling their entry point these days, or even just in a `<script>` tag on the main page, that way you can have `mode = 'firebase'` there and `mode = 'localStorage'` in your test html.

docs/checkit.js
```diff
@@ -773,10 +788,17 @@ var generateUniqueId = function() {
 };
 
 var emptyCalendarState = function(params) {
+    var startDate = moment(params.startDate, "YYYY-MM-DD");
+    var endDate = moment(params.endDate, "YYYY-MM-DD");
+    // Should an error be thrown if the calendar title parameter is an empty string?
```
johntyree: I probably would.

johntyree (changes requested):

docs/checkit.js
```diff
@@ -84,7 +84,12 @@ function CheckIt() {
     
     // We have to connect to Firebase before we can access it.
     // WHEN PAGE LOADS
-    this.initFirebase();
+    if (this.mode === 'firebase') {
+        this.initFirebase();
+    }
+    else {
```
johntyree: I'd probably explicitly check for `'localStorage`' and throw otherwise. That protects you against typos like `new Checkit('firebsae');`.
```
switch (this.mode) {
  case 'firebase': this.initFirebase(); break;
  case 'localStorage': this.initLocalStorage(); break;
  default: throw new Error("Unknown storage mode: '" + this.mode + "'");
}
```

johntyree (changes requested):

docs/checkit.js
```diff
@@ -82,13 +82,11 @@ function CheckIt(mode) {
...
     
-    // We have to connect to Firebase before we can access it.
-    // WHEN PAGE LOADS
-    if (this.mode === 'firebase') {
-        this.initFirebase();
-    }
-    else {
-        this.initLocalStorage();
+    // Initialize storage
+    switch (this.mode) {
```
johntyree: Did you test it?

johntyree (changes requested):

docs/template.html
```diff
@@ -0,0 +1,162 @@
+'<div class="header">\
```
johntyree: You can remove this file now.

docs/checkit.js
```diff
@@ -82,9 +83,12 @@ function CheckIt() {
     this.$signInButton.click(this.signIn.bind(this));
    
     
-    // We have to connect to Firebase before we can access it.
-    // WHEN PAGE LOADS
-    this.initFirebase();
+    // Initialize storage
+    switch (this.mode) {
```
johntyree, line 87: Where is the test for this?

johntyree (changes requested):

johntyree (approved):

PR #141, opened 2017-02-14
Need to modify rules so that I have permission to delete a calendar 

andrayantelo:
from calendars dir in the database.

Ideas I've had so far:
Make a writers dir for the calendars dir. This would kind of be like administrative access to calendars/
Addresses #140

### DIFF
--- a/docs/store.js
+++ b/docs/store.js
@@ -215,12 +215,12 @@ var firebaseCalendarStorage = function(params) {
                     self.endWork();
                 })
                 .catch(function(err) {
-                    console.error('Unable to remove calendar: ' + err);
+                    console.error(`Unable to remove calendar from calendars/ : ` + err);
                     self.endWork();
                 })
             })
             .catch(function(err) {
-                console.error(`Unable to remove calendar from user's allCalendarIds:  err`);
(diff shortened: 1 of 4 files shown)


sample_calendar.json
```diff
@@ -1,78 +1,81 @@
...
-             },
-             "writers" : {
-                 "JULxYGaHuFYlSkXqTC2Fx9KIofp2" : true
-             }
+		},
+		"148461051598386460": {
+			"calendarState": {
+				"startDateString": "20170103",
+				"endDateString": "20170524",
+				"title": "ass",
```
johntyree: Can we title this something other than "ass" now? I open this at work and my calendars are always replaced with "ass" written on the screen in large letters.

docs/store.js
```diff
@@ -155,6 +155,8 @@ var firebaseCalendarStorage = function(params) {
         
         // Store the calendar into allCalendarIds, store calState
         var updates = {};
+        // Should I be updating the allCalendarIds every time I save? saving
```
johntyree: What's expensive is establishing the call to the database. Adding a little extra data probably doesn't matter much.

johntyree (changes requested):

johntyree (approved):

PR #143, opened 2017-02-21
Refactored checkmark click handler and generate checkmarks functions

andrayantelo:
You used to attach the click handler for the checkmarks inside of the month object. I moved it so that it is now something that occurs in the checkit Object because I want access to the store without having to access it from the month object. Previously this happened by passing the checkit object to the calendar object and then passing the calendar object to the month object.
Now no objects get passed to other objects. The click handler gets attached inside of the checkit object. 

I also made generateCheckMarks into a calendar method instead of a month object method. I did this because inside of generateCheckMarks you need to access the checkedDays object which is found in the Calendar object's state and since I am no longer passing the calendar object to month object...

I am wondering if it's okay to have month Object just be three methods (probably will go down to two, because I am planning of getting rid of the removeEmptyWeeks method in the future). The only things you will be able to do with just a month object then is make the html for it, and fill it with the correct information for the month. No checkmarks or anything like that. What do you think?

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -32,9 +32,6 @@ function CheckIt() {
     this.$buildCalendarForm = $('#collapseOne');
     this.$calendarDiv = $('#calendarDiv');
     
-    // The active calendar so that you can use it inside of checkIt.
-    this.calendar;
-    
     var opts = {
       lines: 13 // The number of lines to draw
     , length: 28 // The length of each line
@@ -191,8 +188,8 @@ CheckIt.prototype.displayActiveCalendar = function() {
                .then(function (activeCalendarState) {
                    if (activeCalendarState !==  null) {
                        var state = activeCalendarState;
-                       this.calendar = new Calendar(state);
-                       this.displayCalendar(this.calendar);
+                       calendar = new Calendar(state);
+                       this.displayCalendar(calendar);
                    }
                }.bind(this))
                .catch(function(err) {
@@ -375,16 +372,16 @@ CheckIt.prototype.createCalendar = function() {
         
         //make calendar object
 
-        this.calendar = new Calendar(state);
+        calendar = new Calendar(state);
         
         // Initialize calendar in the storage
-        this.store.initializeCalendar(this.calendar);
+        this.store.initializeCalendar(calendar);
 
         //add calendar to dropdown
-        this.addCalendarToDropdown(this.calendar.state.uniqueId, this.calendar.state.title);
+        this.addCalendarToDropdown(calendar.state.uniqueId, calendar.state.title);
     
         //build the calendar
-        this.buildCalendar(this.calendar);
+        this.buildCalendar(calendar);
         this.collapseBuildMenu(); 
    }
 };
@@ -397,8 +394,8 @@ CheckIt.prototype.loadFromDropdown = function( event ) {
     
     return this.store.loadById(dropdownItemId)
         .then(function(state) {
-            this.calendar = new Calendar(state);
-            this.displayCalendar(this.calendar);
+            calendar = new Calendar(state);
+            this.displayCalendar(calendar);
             this.collapseBuildMenu(); 
         }.bind(this))
         .catch(function() {
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -96,7 +96,7 @@ function CheckIt() {
 
 CheckIt.prototype.addMonth = function() {
     // Add a single month to the calendar
-    
+    // TODO
     //create the month object
     // Need a dateString, possible end Date from cal
     // Maybe grab the last month in the page, and that's how you can get
@@ -142,7 +142,7 @@ CheckIt.prototype.attachCellClickHandler = function(monthObjArray) {
             checkitObj.store.save(checkitObj.calendar);
             
         })
-    }
+    })
    
 };
 
@@ -195,8 +195,8 @@ CheckIt.prototype.displayActiveCalendar = function() {
                .then(function (activeCalendarState) {
                    if (activeCalendarState !==  null) {
                        var state = activeCalendarState;
-                       this.calendar = new Calendar(state, this);
-                       this.displayCalendar(calendar);
+                       this.calendar = new Calendar(state);
+                       this.displayCalendar(this.calendar);
                    }
                }.bind(this))
                .catch(function(err) {
@@ -381,16 +381,16 @@ CheckIt.prototype.createCalendar = function() {
 
         this.calendar = new Calendar(state);
         //add months to checkit's months array
-        this.months = calendar.monthObjects;
+        this.months = this.calendar.monthObjects;
         
         // Initialize calendar in the storage
-        this.store.initializeCalendar(calendar);
+        this.store.initializeCalendar(this.calendar);
 
         //add calendar to dropdown
-        this.addCalendarToDropdown(calendar.state.uniqueId, calendar.state.title);
+        this.addCalendarToDropdown(this.calendar.state.uniqueId, this.calendar.state.title);
     
         //build the calendar
-        this.buildCalendar(calendar);
+        this.buildCalendar(this.calendar);
         this.collapseBuildMenu(); 
    }
 };
@@ -403,8 +403,8 @@ CheckIt.prototype.loadFromDropdown = function( event ) {
     
     return this.store.loadById(dropdownItemId)
         .then(function(state) {
-            this.calendar = new Calendar(state, this);
-            this.displayCalendar(calendar);
+            this.calendar = new Calendar(state);
+            this.displayCalendar(this.calendar);
             this.collapseBuildMenu(); 
         }.bind(this))
         .catch(function() {
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -57,6 +57,7 @@ function CheckIt() {
     
     this.spinner = new Spinner();
     
+    // Click handlers for the DOM
     this.$clearButton.click(this.clearForm.bind(this));
     this.$createButton.click(this.createCalendar.bind(this));
     this.$calendarDropdown.on('click', 'li', this.loadFromDropdown.bind(this));
@@ -96,7 +97,31 @@ CheckIt.prototype.addMonth = function() {
     // the month and year for the new month, end it on whatever endDate
     // user picked for the previous month (then previous month would need to be
     // filled all the way too. 
-    //newMonth = new Month(
+    //var newMonth = new Month(
+    
+    // Get the active Calendar information because you need to know what the 
+    // last month is as well as the endDate.
+    this.store.getActive()
+        .then(function (activeCalendarId) {
+            this.store.loadById(activeCalendarId)
+                .then(function (activeCalendarState) {
+                   if (activeCalendarState !==  null) {
+                       var state = activeCalendarState;
+                       console.log(state);
+                   }
+               }.bind(this))
+               .catch(function(err) {
+                   console.error("Could not load calendar " + err);
+                   this.store.removeActive();
+                   this.uncollapseBuildMenu();
+               }.bind(this));
+           
+       }.bind(this))
+       
+       .catch(function () {
+           console.log("Could not load current active calendar");
+           this.uncollapseBuildMenu();
+       }.bind(this));
     
     
     //this.attachCellClickHandler([newMonth]);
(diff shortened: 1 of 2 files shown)


docs/checkit.js
```diff
@@ -91,6 +88,42 @@ function CheckIt() {
 
 };
 
+CheckIt.prototype.attachCellClickHandler = function(calObj) {
+    
+    var checkitObj = this;
+    this.$calendarDiv.find('.cell').click(function(event) {
```
johntyree: Note that this is going to have to change or it will attach additional handlers when you try to add months to the page.

docs/checkit.js
```diff
@@ -91,6 +88,42 @@ function CheckIt() {
...
+        
+        //HARDCODED FOR NOW
+        
+        var $monthDiv = $(`#${self.monthId}`);
+        var boxId = $(this).attr('id');
+        //if the boxId is not checked (as in, the value is not inside of checkedDays
```
johntyree: this comment belongs down on 107

docs/checkit.js
```diff
@@ -91,6 +88,42 @@ function CheckIt() {
...
+        
+        if (calObj.state.checkedDays === undefined) {
+            calObj.state.checkedDays = {};
+        }
+        
+        if (calObj.state.checkedDays[boxId] === undefined) {
+            //add it to checkedDays
+            calObj.state.checkedDays[boxId] = 1;
+            //then add a checkmark
+            $(this).find('.element').removeClass("hidden");
```
johntyree: `.element` is not a very informative name.

docs/checkit.js
```diff
@@ -865,10 +779,35 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
     monthObjectsArray.forEach (function(monthObj) {
         monthObj.fillMonthDiv();
         monthObj.removeEmptyWeeks();
-        monthObj.attachClickHandler();
-        monthObj.generateCheckmarks();
+        //monthObj.attachClickHandler();
```
johntyree: Remove these

docs/checkit.js
```diff
@@ -865,10 +779,35 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+    if (self.state.checkedDays === undefined) {
+        console.log("No days are checked");
+        return;
+    }
+    
+    $('#calendarDiv').find('.cell').each( function() 
+    {
+        var boxId = $(this).attr('id');
+        
+        if (self.state.checkedDays[$(this).attr('id')])
```
johntyree: `[boxId]`

docs/checkit.js
```diff
@@ -865,10 +779,35 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+    // object.
+    
+    //checkedDays is an object that contains a date that points to 1 or 0
+    var self = this;
+    if (self.state.checkedDays === undefined) {
+        console.log("No days are checked");
+        return;
+    }
+    
+    $('#calendarDiv').find('.cell').each( function()
```
johntyree: Booooooooooo.... :-1: 

K&R style braces please!

```
block {
}
```
not
```
block
{
}
```

docs/checkit.js
johntyree: Also braces

johntyree (changes requested):

docs/checkit.js
andrayantelo: Fixed.

docs/checkit.js
andrayantelo: Fixed.

docs/checkit.js
andrayantelo: Removed.

docs/checkit.js
andrayantelo: Changed to checkmark.

docs/checkit.js
andrayantelo: Fixed.

docs/checkit.js
```diff
@@ -91,6 +94,58 @@ function CheckIt() {
...
+    
+    //this.attachCellClickHandler([newMonth]);
+};
+
+CheckIt.prototype.attachCellClickHandler = function(monthObjArray) {
+    // Takes an array of month objects (the array can be of length 1)
+    var checkitObj = this;
+    
+    monthObjArray.forEach( function(monthObj) {
+        var $monthDiv = $(`#${monthObj.monthId}`);
```
johntyree: prefer normal strings over backticks unless you're getting a lot of bang for your buck.

```
`${user}-${host}-${ip}.data.log`
```
for example

docs/checkit.js
```diff
@@ -865,10 +803,32 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
         monthObj.fillMonthDiv();
         monthObj.removeEmptyWeeks();
-        monthObj.attachClickHandler();
-        monthObj.generateCheckmarks();
     });
 };
 
 
+Calendar.prototype.generateCheckmarks = function() {
+    // Toggles the hidden class between the children of the div class="cell"
```
johntyree: This comment is inaccurate.

docs/checkit.js
```diff
@@ -507,6 +564,8 @@ CheckIt.prototype.buildCalendar = function(calendarObject) {
     
     calendarObject.generateEmptyCalendar(calendarObject.monthObjects);
     calendarObject.fillCalendar(calendarObject.monthObjects);
+    this.attachCellClickHandler(calendarObject.monthObjects);
+    calendarObject.generateCheckmarks();
```
johntyree: Are we planning on moving this dom manipulation code onto the checkit object?

johntyree (changes requested):

docs/checkit.js
andrayantelo: Fixed that particular line.

docs/checkit.js
andrayantelo: Fixed.

docs/checkit.js
andrayantelo: I suppose so. :)

docs/checkit.js
```diff
@@ -91,6 +88,80 @@ function CheckIt() {
...
+        .then(function (activeCalendarId) {
+            this.store.loadById(activeCalendarId)
+                .then(function (activeCalendarState) {
+                   if (activeCalendarState !==  null) {
+                       var state = activeCalendarState;
+                       console.log(state);
+                       // Create a calendar so that you can manipulate this calendar State
+                       var calendar = new Calendar(state);
+                       // Add a month to this calendar, function has the same name
+                       // as the one we are currently in. Bad practice?
```
johntyree: It's not bad practice if it clearly describes what you are trying to do. I think it's probably fine. If you don't like it, perhaps you could expand the names to be more precise?

docs/checkit.js
```diff
@@ -91,6 +88,80 @@ function CheckIt() {
...
+                       calendar.addMonth();
```
johntyree: Rather than making an assumption about how / where the calendar object will add the month, just have it return the month that gets added. Then you can pass it to `attachCellClickHandler`.

docs/checkit.js
```diff
@@ -91,6 +88,80 @@ function CheckIt() {
...
+                       this.attachCellClickHandler(calendar, [calendar.lastMonth]);
+                   }
+               }.bind(this))
+               .catch(function(err) {
+                   console.error("Could not load calendar " + err);
+                   this.store.removeActive();
```
johntyree: Add a comment about why you're doing this.

docs/checkit.js
```diff
@@ -91,6 +88,80 @@ function CheckIt() {
...
+        var $monthDiv = $('#' + monthObj.monthId);
+        $monthDiv.find('.cell').click(function(event) {
+            
+            var boxId = $(this).attr('id');
+            
+            
+            if (calObj.state.checkedDays === undefined) {
+                calObj.state.checkedDays = {};
+            }
+            //if the boxId is not checked (as in, the value is not inside of checkedDays
```
johntyree: This is somewhat difficult to follow. Maybe, "We only store data for days with checkmarks. Therefore, unchecked days are undefined."

docs/checkit.js
```diff
@@ -140,7 +211,7 @@ CheckIt.prototype.displayActiveCalendar = function() {
                .then(function (activeCalendarState) {
                    if (activeCalendarState !==  null) {
                        var state = activeCalendarState;
-                       var calendar = new Calendar(state, this);
+                       calendar = new Calendar(state);
```
johntyree: `var` ?

docs/checkit.js
```diff
@@ -324,7 +395,7 @@ CheckIt.prototype.createCalendar = function() {
         
         //make calendar object
 
-        var calendar = new Calendar(state, this);
+        calendar = new Calendar(state);
```
johntyree: Why do these all have `var` removed?

docs/checkit.js
```diff
@@ -346,7 +417,7 @@ CheckIt.prototype.loadFromDropdown = function( event ) {
     
     return this.store.loadById(dropdownItemId)
         .then(function(state) {
-            var calendar = new Calendar(state, this);
+            calendar = new Calendar(state);
```
johntyree: `var`

docs/checkit.js
```diff
@@ -817,10 +770,26 @@ var Calendar = function(state, checkitObj) {
     //user wants to track
     self.numberOfMonths = self.endDate.diff(self.startDate, 'months', true);
     self.monthObjects = self.generateMonthObjects();
-    self.checkit = checkitObj;
-    
+    // Last month to be added, this is so that when you need to attach the
+    // click handler to this month, you don't do it to the entire calendar.
+    self.lastMonth = null;
```
johntyree: You probably don't need this. At the very least you can always just grab it manually from the `monthObjects` array.

docs/checkit.js
```diff
@@ -817,10 +770,26 @@ var Calendar = function(state, checkitObj) {
...
+    // Update the self.lastMonth property
+    self.lastMonth = new Month(self.endDate.format("YYYYMMDD"));
+    // Update the endDateString in the calendarState
+    self.state.endDateString = self.endDate.format("YYYYMMDD");
+    // Update the self.monthObjects so that it includes the new month
+    self.monthObjects = self.generateMonthObjects();
+};
+
     
 Calendar.prototype.generateMonthObjects = function() {
```
johntyree: If this function took `startDate` and `endDate` as parameters you could also use it to create a single month as needed.

docs/checkit.js
```diff
@@ -865,10 +834,32 @@ Calendar.prototype.fillCalendar = function(monthObjectsArray) {
...
+    // object.
+    
+    //checkedDays is an object that contains a date that points to 1 or 0
+    var self = this;
+    if (self.state.checkedDays === undefined) {
+        console.log("No days are checked");
+        return;
+    }
+    
+    $('#calendarDiv').find('.cell').each( function() {
```
johntyree, line 850: If this function took this element as a parameter, you could use it to initialize single months or the whole calendar as needed.

johntyree (changes requested):

docs/checkit.js
andrayantelo: Everything about this function is changed in my other PR.

andrayantelo: I mean we can revisit this in the other PR.

docs/checkit.js
andrayantelo: Probably bad copy pasting.

docs/checkit.js
```diff
@@ -782,32 +778,33 @@ Calendar.prototype.addMonth = function() {
     
     // Update self.endDate so that it occurs one month later.
     self.endDate = self.endDate.add(1, 'months');
-    // Update the self.lastMonth property
-    self.lastMonth = new Month(self.endDate.format("YYYYMMDD"));
     // Update the endDateString in the calendarState
     self.state.endDateString = self.endDate.format("YYYYMMDD");
     // Update the self.monthObjects so that it includes the new month
-    self.monthObjects = self.generateMonthObjects();
+    self.monthObjects = self.generateMonthObjects(self.startDate, self.endDate);
```
johntyree: I was thinking something like..

```
var oldEndDate = Moment(self.endDate);
self.endDate.add(1, 'months');  // I think this just mutates the object instead of copying it and returning the new one right?
var newMonth = self.generateMonthObjects(oldEndDate, self.endDate);
self.monthObjects = self.monthObjects.push(newMonth);
return newMonth;
```

But regenerating all the month objects is probably fine. There aren't that many of them and it makes it harder to accidentally have stale state hanging around.

johntyree (approved):

docs/checkit.js
johntyree: sure

PR #150, opened 2017-02-22
Made generateEmptyCalendar and fillCalendar into Checkit methods

andrayantelo:
This branch I created from master, and then I merged the checkmarkClickHandler branch into it, which is the branch where I worked on making the checkmark click handler something that the checkit object handles. 

So, now I can generate an empty calendar, fill it, and attach the click handler in the checkit object. Am missing the function that generates in checkmarks (like when you have saved checkmarks), and the function that removes empty weeks. Still need to make those two into checkit object methods. 

This PR is already huge.. so, how should I proceed? Wait til you approve this one? Does the other PR even matter anymore (The one where I made the click Handler into something that is handled by the checkit object). @johntyree

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -354,6 +354,98 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
     }
 };
 
+CheckIt.prototype.generateEmptyCalendar = function(calObj) {
+    // Generate the html for an empty calendar of the calendar you want to 
+    // display.
+    
+    var calendarDiv = this.$calendarDiv;
+    
+    console.log("Running checkit's generateEmptyCalendar");
+    // Add the title of the calendar
+    calendarDiv.append('<div id="calendarTitleHeading"> <h1 class="page-header text-center">' +
+                        calObj.state.title + '</h1></div>');
+                        
+
+    calObj.monthObjects.forEach (function(monthObj, index) {
+
+        var yearHeader = "<div class='page-header text-center'>" +
+                         "<h2 id='yearHeader'>" + monthObj.monthYear + "</h2>" +
+                         "</div>";
+        
+        //the div ID is the monthID
+            
+        calendarDiv.append('<div class="monthframe" id=' + monthObj.monthId + '></div>');
+        
+        if (self.monthIndex === 0) {
+            $('#' + monthObj.monthId).append(yearHeader);
+        }
+        $('#' + monthObj.monthId).append($('#template').html());
+            
+    });
+    
+};
+
+CheckIt.prototype.fillCalendar = function(calObj) {
+    // Fill an empty calendar with appropriate calendar data.
+    console.log("Running checkit's fillCalendar method");
+    calObj.monthObjects.forEach (function(monthObj) {
+
+        var $monthId = $('#'+ monthObj.monthId);
+        
+        //self.clearMonthDiv();  <-- Do I need this?
+        
+        $monthId.find(".month-year").empty();
+        $monthId.find(".month-year").append(monthObj.monthName + " " + monthObj.monthYear);
+        
+        //go through each td and fill in correct day number
+        $monthId.find($('.week')).find('td').each( function(indexOfTableTd) {
+            //the indexOfTableSquare is where we are currently on the month table
+            //which td are we in, from 0 to 42, because there are 6 rows
+            // or 7 columns 
+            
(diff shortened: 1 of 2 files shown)


docs/checkit.js
```diff
@@ -354,12 +356,92 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
-CheckIt.prototype.checkSignedInWithMessage = function() {
-    // Return true if the user is signed in Firebase
-    if (this.auth.currentUser) {
-        return true;
-    }
+CheckIt.prototype.generateEmptyCalendar = function(calObj) {
+    // Generate the html for an empty calendar of the calendar you want to 
+    // display.
+    
+    var calendarDiv = this.$calendarDiv;
```
johntyree: The point of prefixing variables with `$` is so you know whether or not it's already jquery selector.

docs/checkit.js
```diff
@@ -354,12 +356,92 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
+    
+    // Add the title of the calendar
+    calendarDiv.append('<div id="calendarTitleHeading"> <h1 class="page-header text-center">' +
+                        calObj.state.title + '</h1></div>');
+                        
+
+    calObj.monthObjects.forEach (function(monthObj, index) {
+
+        var yearHeader = "<div class='page-header text-center'>" +
+                         "<h2 id='yearHeader'>" + monthObj.monthYear + "</h2>" +
```
johntyree: My understanding is that `id` attributes should be unique across the page.

docs/checkit.js
```diff
@@ -354,12 +356,92 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
+                         "</div>";
+        
+        //the div ID is the monthID
+            
+        calendarDiv.append('<div class="monthframe" id=' + monthObj.monthId + '></div>');
+        
+        if (self.monthIndex === 0) {
+            $('#' + monthObj.monthId).append(yearHeader);
```
johntyree: Why create the year header outside of the this conditional?

docs/checkit.js
```diff
@@ -354,12 +356,92 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
+
+CheckIt.prototype.fillCalendar = function(calObj) {
+    // Fill an empty calendar with appropriate calendar data.
+
+    calObj.monthObjects.forEach (function(monthObj) {
+        
+        var $monthId = $('#'+ monthObj.monthId);
+        
+        $monthId.find(".month-year").empty();
+        $monthId.find(".month-year").append(monthObj.monthName + " " + monthObj.monthYear);
```
johntyree: Are these two lines together are the same as `$monthId.find(".month-year").text(...)`?

docs/checkit.js
```diff
@@ -354,12 +356,92 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
+        
+        // Go through each td and fill in correct day number
```
johntyree: What is a td here?

docs/checkit.js
```diff
@@ -576,10 +658,11 @@ CheckIt.prototype.buildCalendar = function(calendarObject) {
     //this function assumes the calendarObject already has it's
     //state updated with the correct information. 
     
-    calendarObject.generateEmptyCalendar(calendarObject.monthObjects);
-    calendarObject.fillCalendar(calendarObject.monthObjects);
-    this.attachCellClickHandler(calendarObject, calendarObject.monthObjects);
-    calendarObject.generateCheckmarks();
+    this.generateEmptyCalendar(calendarObject);
+    this.fillCalendar(calendarObject);
+    //calendarObject.fillCalendar(calendarObject.monthObjects);
```
johntyree: remove commented code

johntyree (changes requested):

docs/checkit.js
andrayantelo: Don't actually need that id, since bootstrap classes take care of the CSS. Removed it.

docs/checkit.js
andrayantelo: No reason. Moved it.

docs/checkit.js
andrayantelo: the element <td></td> which has the .cell div in it which contains the div that has the day number in it and the checkmark div.

johntyree (approved):

johntyree (approved):

PR #153, opened 2017-02-25
Made generateCheckmarks into a checkit method

andrayantelo:


docs/checkit.js
```diff
@@ -440,6 +440,27 @@ CheckIt.prototype.fillCalendar = function(calObj) {
     })
 };
 
+CheckIt.prototype.generateCheckmarks = function(calObj, div) {
+    // Display saved checkmarks on calendar
+    // div is which div do you want to look through for checkmarks
+    
+    var $div = div;
```
johntyree: Why not just name the variable `$div` in the function signature?

johntyree (changes requested):

johntyree (approved):

PR #154, opened 2017-02-25
Remove empty weeks function is now checkit method.

andrayantelo:


docs/checkit.js
```diff
@@ -460,6 +460,22 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $div) {
      })
 };
 
+CheckIt.prototype.removeEmptyWeeks = function(calObj, $div) {
+    // Remove empty weeks from the calendar
+        
+    $div.find('.month').find('.week').each( function(index) {
+        var counter = 0;
+        $(this).find('td').each(function(td) {
+            if ($(this).children().hasClass("nill"))
```
johntyree: Not that it matters, but "nill" is usually spelled "nil."

docs/checkit.js
```diff
@@ -460,6 +460,22 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $div) {
      })
 };
 
+CheckIt.prototype.removeEmptyWeeks = function(calObj, $div) {
```
johntyree: `$div` could have a more descriptive name. This function can't take just any div. It has to be a `$calendarDiv` that contains `.month` and `.week` etc.

docs/checkit.js
```diff
@@ -460,6 +460,22 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $div) {
...
+        var counter = 0;
+        $(this).find('td').each(function(td) {
+            if ($(this).children().hasClass("nill")) 
+            counter +=1;
+        })
+        if (counter === 7) 
+        {
+            $(this).remove();
+        }
+    })
```
johntyree: This is a clever way to figure out if the week should be removed or not, but you can also just ask jquery how many things have the `nil` class.

Try something like `$(this).children('.nil').length`

johntyree (changes requested):

docs/checkit.js
andrayantelo: Gosh darn it. It matters to me.

docs/checkit.js
andrayantelo: Changed name to $calendarDiv

docs/checkit.js
```diff
@@ -460,17 +458,11 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $div) {
...
         
-    $div.find('.month').find('.week').each( function(index) {
-        var counter = 0;
-        $(this).find('td').each(function(td) {
-            if ($(this).children().hasClass("nill")) 
-            counter +=1;
-        })
-        if (counter === 7) 
-        {
+    $calendarDiv.find('.month').find('.week').each( function(index) {
```
johntyree: These are CSS selectors, right? I don't think you should have to do multiple calls in a row. `find('.month > .week')`

docs/checkit.js
```diff
@@ -460,17 +458,11 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $div) {
...
+        if ($(this).find('td').children('.nil').length === 7) {
```
johntyree: `.find("td > .nil")`

johntyree (changes requested):

johntyree (approved):

PR #156, opened 2017-02-27
Fixed selector syntax

andrayantelo:


docs/checkit.js
```diff
@@ -460,8 +460,8 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $calendarDiv) {
 
 CheckIt.prototype.removeEmptyWeeks = function(calObj, $calendarDiv) {
     // Remove empty weeks from the calendar
-        
-    $calendarDiv.find('.month > .week').each( function(index) {
+    
+    $calendarDiv.find('.month > tbody > .week').each( function(index) {
```
johntyree: Just use `'.month .week'` probably. You don't want this to break just because you change how the calendar part is structured in html

johntyree (changes requested):

docs/checkit.js
andrayantelo: > is for direct children. .week is not a direct child of .month

andrayantelo: ">" is for direct children.

andrayantelo: Could give tbody a class?

andrayantelo: .monthBody? Then it would be .month > .monthBody > .week

andrayantelo: Okay, if I just leave it as '.month . week' it works. This is how you select a descendant. (not necessarily a direct child).

johntyree (approved):

PR #157, opened 2017-02-27
Different CSS for current day

andrayantelo:


johntyree (changes requested):
This will be out of date in at most 24 hours. It should check the date again and update as needed.

Or maybe just refresh at 12:01 am localtime? I don't know.

johntyree (approved):

PR #166, opened 2017-03-05
Writing tests for checkit object

andrayantelo:


tests/tests.js
```diff
@@ -145,7 +145,68 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
+    this.calendar = new Calendar(this.state);
+    this.checkit = new CheckIt('localStorage');
+}
+
+});
+
+function selectorNameTest(testName, selectorName, expected) {
+    QUnit.test(testName, function(assert) {
+        assert.expect(1);
+        assert.equal(selectorName.selector, expected);
```
johntyree, line 279: This is good to check, but it doesn't protect you against something changing on the page. For example, suppose you refactor how the calendar works and need to change your date form selectors. This test will still pass but the selector won't *work*.

Is there a good way to test that each of the selectors actually selects an element?

johntyree (changes requested):

docs/checkit.js
```diff
@@ -381,24 +326,23 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
-
     calObj.monthObjects.forEach (function(monthObj, index) {
         
         //the div ID is the monthID
         $calendarDiv.append('<div class="monthframe" id=' + monthObj.monthId + '></div>');
         
-        if (self.monthIndex === 0) {
-            var yearHeader = "<div class='page-header text-center'>" +
-                "<h2>" + monthObj.monthYear + "</h2>" +
+        if (monthObj.monthIndex === 0 || monthObj.isFirst) {
```
johntyree: Is `monthObj.isFirst` the same as `index === 0` ?

johntyree: Do you need it elsewhere?

johntyree (changes requested):

docs/checkit.js
andrayantelo: isFirst means it is the very first month in the calendar the user built. index === 0 means it is January, and January can appear multiple times in the calendar.

docs/checkit.js
```diff
@@ -92,43 +69,11 @@ function CheckIt(mode) {
...
-       }.bind(this))
-       
-       .catch(function () {
-           console.log("Could not load current active calendar");
-           this.uncollapseBuildMenu();
-       }.bind(this));
-    
-    
-    
+    // Add a month to this calendar
```
johntyree: This comment seems wrong

docs/checkit.js
johntyree: `index`, not `monthObj.index`. `index` is coming from the `monthObjects.forEach` loop above.

johntyree (changes requested):

docs/checkit.js
```diff
@@ -676,21 +680,28 @@ CheckIt.prototype.clearPage = function() {
...
     
-    this.$calendarDiv.find('.today').removeClass("today");
-    
-    var today = moment();
+    var today = moment(currentDay, "YYYYMMDD") || moment();
     var todayId = moment({"year":today.year(), "month":today.month(), "day": today.date()}).format("YYYYMMDD");
-    
-    this.$calendarDiv.find('#' + todayId).addClass("today");
+    // If current day is not actually an ACTIVE day in the calendar, don't add it
+    this.$calendarDiv.find('.activeDay').each( function(index, activeDay) {
```
johntyree: Try 
```
this.$calendarDiv.find('.activeDay')
                 .children('.cell')
                 .filter('#' + todayId)
                 .addClass('currentDay');
```

docs/store.js
```diff
@@ -361,8 +361,8 @@ var LocalCalendarStorage = function(params) {
     
     var toKey = function(id) {
         //make a key out of a uniqueId
-        
-        var key = prefix + "_" + id;
+        var key = id;
```
johntyree: `var key = prefix ? prefix + "_" + id : id;`

tests/tests.js
```diff
@@ -193,13 +183,15 @@ QUnit.test("generateEmptyCalendar test", function( assert ) {
         assert.equal($(this).html(), daysOfWeek[index]);
     });
     
-    //TODO test with a calendar with more than one month to make sure
-    // the second month does not get a yearHeader.
-    
+    this.checkit.addMonth(this.calendar);
+    $calendarDiv.empty();
```
johntyree: If you need to do this, then you probably should start a second test and give them better names.

tests/tests.js
```diff
@@ -326,104 +312,311 @@ QUnit.test("Initialize CheckIt test", function( assert ) {
...
+  }
+});
+
+
+QUnit.test("attachCheckMarkClickHandler test", function( assert ) {
+    assert.expect(6);
+    this.checkit.attachCheckmarkClickHandler(this.calendar,
+        this.calendar.monthObjects);
+    
+    assert.equal(this.$calendarDiv.find('#20171').find('.hidden').length, 7);
```
johntyree: If you use a better selector you only need one call to `find`.

tests/tests.js
```diff
@@ -326,104 +312,311 @@ QUnit.test("Initialize CheckIt test", function( assert ) {
...
+    // Check valentine's day
+    this.$calendarDiv.find('#20170214').trigger('click');
+    assert.equal(this.$calendarDiv.find('#20171').find('.hidden').length, 6);
```
johntyree: Same here. One `find` call.

tests/tests.js
```diff
@@ -326,104 +312,311 @@ QUnit.test("Initialize CheckIt test", function( assert ) {
...
-    assert.expect(2);
+QUnit.test("findCurrentDay test", function( assert ) {
+    assert.expect(3);
     
-    var done = assert.async();
-    var fixture = this.fixture;
-    var testmonth = this.testmonth;
+    var currentDay = "20170214";
+    assert.equal(this.$calendarDiv.find('#20170214').attr('class'), 'cell');
+    this.checkit.findCurrentDay();
```
johntyree: What does this function do? The name suggests that it's going to e.g. locate the dom element corresponding to today and return it or something.

When you write functions that are going to mutate state, it's important to make sure that it's obvious that that's what's happening.

tests/tests.js
```diff
@@ -326,104 +312,311 @@ QUnit.test("Initialize CheckIt test", function( assert ) {
...
 
-QUnit.test("attachClickHandler test", function( assert ) {
+QUnit.module( "Checkit DOM manipulation tests", {
+  beforeEach: function() {
+    // prepare something before each test
+    this.params = {startDate: "2017-02-14" , endDate: "2017-02-20" , calendarTitle: "Test Calendar"};
+    this.state = emptyCalendarState(this.params);
+    this.calendar = new Calendar(this.state);
+    this.checkit = new CheckIt('localStorage');
+    this.fixture = $('#qunit-fixture');
```
johntyree: `$fixture`

johntyree (changes requested):

tests/tests.js
```diff
@@ -145,127 +144,480 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
+    
+    var daysOfWeek = ['S', 'M', 'T', 'W', 'T', 'F', 'S'];
+    $calendarDiv.find('#dayNames td').each(function(index, td) {
+        assert.equal($(this).html(), daysOfWeek[index]);
+    });
+    
+    this.checkit.addMonth(this.calendar);
+    $calendarDiv.empty();
+    this.checkit.generateEmptyCalendar(this.calendar, $calendarDiv);
+    assert.notOk($calendarDiv.find('20172').find('#yearHeader').text(),
```
johntyree: Combine the calls to `find`

tests/tests.js
```diff
@@ -145,127 +144,480 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
+
+
+});
+
+function fillMonth(testName, monthObj, $div, expected) {
+    QUnit.test(testName, function(assert) {
+        assert.expect(expected);
+        assert.notOk(jQuery.isEmptyObject(monthObj.dayIndex),
+            "dayIndex object for month is not empty");
+        $div.find('#' + monthObj.monthId).find('.daynumber').each( function(index) {
```
johntyree: Combine `find` calls

tests/tests.js
```diff
@@ -145,127 +144,480 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
-        assert.equal($(this).find('.cell').children().length, 30, "Total number of children of each .cell div");
-        assert.ok($(this).find('.daynumber').text(), ".daynumber divs have text in them");
-        
-        $(this).find('.daynumber').each( function(index) {
-            assert.equal($(this).html(), index + 14, "Correct daynumbers");
-            });
-            
-        $(this).find('.cell').each(function (index) {
-            assert.ok($(this).find('.element').hasClass("hidden"), "Check that .element divs have a hidden class");
+        $div.find('#' + monthObj.monthId).find('.checkmark').each(function (index) {
```
johntyree: Also here

tests/tests.js
```diff
@@ -145,127 +144,480 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
+        this.calendar.monthObjects);
+    
+    assert.equal(this.$calendarDiv.find('#20171 .hidden').length, 7);
+    // Check valentine's day
+    this.$calendarDiv.find('#20170214').trigger('click');
+    assert.equal(this.$calendarDiv.find('#20171 .hidden').length, 6);
+    assert.equal(this.calendar.state.checkedDays['20170214'], 1);
+    // Uncheck valentine's day
+    this.$calendarDiv.find('#20170214').trigger('click');
+    assert.equal(this.$calendarDiv.find('#20171').find('.hidden').length, 7);
```
johntyree: and here

tests/tests.js
```diff
@@ -145,127 +144,480 @@ QUnit.test("generateMonthObjects calendar method test", function(assert) {
...
+    this.$calendarDiv.find('#20170214').trigger('click');
+    // Clear calendar, and build again
+    this.$calendarDiv.empty();
+    this.checkit.generateEmptyCalendar(this.calendar, this.$calendarDiv);
+    this.checkit.fillCalendar(this.calendar);
+    this.checkit.attachCheckmarkClickHandler(this.calendar,
+        this.calendar.monthObjects);
+    this.checkit.generateCheckmarks(this.calendar, this.$calendarDiv);
+    // Make sure valentine's day is checked on the DOM
+    assert.equal(this.$calendarDiv.find('#20170214').find('.checkmark')
```
johntyree: and here

johntyree (changes requested):

PR #168, opened 2017-03-06
CheckIt Tests Part II

andrayantelo:
Wrote a few tests but for the most part ended up refactoring checkit methods.


docs/checkit.js
```diff
@@ -438,22 +441,175 @@ CheckIt.prototype.removeEmptyWeeks = function(calObj, $calendarDiv) {
...
-    this.$fullForm[0].reset();
+};
+
+CheckIt.prototype.removeFieldError = function($id, $srId) {
+    // Removes error state CSS on form field 
+    // $id is the selector for the div with class .form-group that
+    // surrounds the input field with the error
+    // $srId is the selector for the screen reader element
+    $id.removeClass('has-error has-feedback');
+    $srId.addClass('hidden');
```
johntyree, line 486: No aria here?

docs/checkit.js
```diff
@@ -438,22 +441,175 @@ CheckIt.prototype.removeEmptyWeeks = function(calObj, $calendarDiv) {
...
+};
+
+CheckIt.prototype.addHelpBlock = function($helpBlock) {
```
johntyree, line 489: Looks like these four add/remove functions can be combined to just two show and hide functions.

docs/checkit.js
```diff
@@ -438,22 +441,175 @@ CheckIt.prototype.removeEmptyWeeks = function(calObj, $calendarDiv) {
...
+    var $helpBlock = $formGroup.find('#helpBlock');
+    var $fiveYears = $formGroup.find('#fiveYears');
+    var difference = endDate.diff(startDate.format("YYYY-MM-DD"), 'years', true);
+    
+    if (startDate.isBefore(endDate)) {
+        // startDate is before endDate so errors can be removed.
+        this.removeFieldError($formGroup, $sr);
+        this.removeHelpBlock($helpBlock);
+        // If there are more than 5 years between the dates return false for invalid
+        if (difference < 5) {
```
johntyree, line 552: Why have this?

docs/checkit.js
```diff
@@ -667,17 +725,17 @@ CheckIt.prototype.buildCalendar = function(calendarObject) {
...
-    this.clearPage();
+    this.clearPage(this.$calendarDiv);
     this.buildCalendar(calendarObj);
     this.store.setActiveById(calendarObj.state.uniqueId);
     this.store.save(calendarObj);
 };
 
-CheckIt.prototype.clearPage = function() {
+CheckIt.prototype.clearPage = function($div) {
     // Remove all divs from page except #template
```
johntyree: This comment is not accurate.

tests/tests.js
```diff
@@ -439,79 +439,275 @@ QUnit.test("hideLoadingWheel test", function( assert ) {
...
 });
 
-QUnit.test("clearForm test", function( assert ) {
-    assert.expect(0);
-});
 
 QUnit.test("addCalendarToDropdown test", function( assert ) {
-    assert.expect(0);
+    assert.expect(2);
+    this.checkit.addCalendarToDropdown("1234", "Hello", this.$dropdown);
```
johntyree: does this still work if there are other entries in the dropdown already?

tests/tests.js
```diff
@@ -439,79 +439,275 @@ QUnit.test("hideLoadingWheel test", function( assert ) {
...
+    assert.expect(1);
+    
+    this.$calendarDiv.append(`<div id="testCollapse" class="collapse"></div>`);
+    
+    var $testCollapse = this.$calendarDiv.find('#testCollapse');
+    assert.equal($testCollapse.css('display'), 'none');
+    this.checkit.showForm($testCollapse);
+    setTimeout(function(){ 
+        assert.equal($testCollapse.css('display'), 'block');
+        }, 350);
```
johntyree, line 540: This works kind of, but it's basically a race condition. What happened to using the transitionend event?

johntyree (approved):
It was pretty hard to slog through reading these tests.

johntyree (changes requested):
Accidental approval.

docs/checkit.js
andrayantelo: What do you mean? srId is the selector for the screen reader element.

docs/checkit.js
andrayantelo: so that people don't make a calendar that is 500 years long. Or a million.

docs/checkit.js
andrayantelo: Removed.

tests/tests.js
andrayantelo: It wouldn't 'hear' the transitionend event, which is why i switched to using setTimeout.

docs/checkit.js
andrayantelo: Didn't do this because of reasons we talked about.

tests/tests.js
andrayantelo: Added testing for that case.

tests/tests.js
johntyree: That's odd. You were attaching the event listener to `$testCollapse` I assume?

docs/checkit.js
```diff
@@ -732,7 +732,7 @@ CheckIt.prototype.displayCalendar = function(calendarObj) {
 };
 
 CheckIt.prototype.clearPage = function($div) {
-    // Remove all divs from page except #template
+    // Remove all calendar related divs from page
```
johntyree: this is still wrong. It removes the checkit related divs from inside `$div`. The rest are untouched.

tests/tests.js
```diff
@@ -463,10 +463,15 @@ QUnit.module( "CheckIt tests dropdown", {
 
 
 QUnit.test("addCalendarToDropdown test", function( assert ) {
-    assert.expect(2);
+    assert.expect(3);
     this.checkit.addCalendarToDropdown("1234", "Hello", this.$dropdown);
     assert.equal(this.$dropdown.children().attr('id'), '1234');
     assert.equal(this.$dropdown.children().text(), "Hello");
+    
+    // Check with items in dropdown already.
```
johntyree: I meant the previous asserts with something in the dropdown already. I suggest.

1. Add made up thing to dropdown
2. Check that 'hello' whatever isn't there.
3. Add "1234", "Hello" to dropdown.
4. Test that it *is* there.

johntyree (changes requested):

docs/checkit.js
andrayantelo: Fixed.

tests/tests.js
andrayantelo: Fixed.

tests/tests.js
```diff
@@ -472,6 +472,12 @@ QUnit.test("addCalendarToDropdown test", function( assert ) {
     this.$dropdown.append("<li>Hello World</li>");
     this.checkit.addCalendarToDropdown("101", "Cal", this.$dropdown);
     assert.equal(this.$dropdown.children().length, 3);
+    
+    this.$dropdown.empty()
```
johntyree: Why do you clear it? Now I don't understand what this is supposed to be doing.

johntyree (approved):

johntyree (changes requested):
Accidentally approved again. Github really needs to make this less annoying to use correctly.

tests/tests.js
andrayantelo: Fixed.

tests/tests.js
andrayantelo: If I remember correctly, yes. This was at the beginning of this week.

johntyree (approved):

PR #169, opened 2017-03-10
Store tests

andrayantelo:


docs/checkit.js
```diff
@@ -235,13 +237,13 @@ CheckIt.prototype.signOut = function() {
 };
 
 // Triggers when there is a change in the storage.
-CheckIt.prototype.onActivityChanged = function(activeCalls) {
+CheckIt.prototype.onActivityChanged = function(activeCalls, id) {
     // Will Manipulate the DOM to show the loading wheel or to hide it.
     // Passing storageObj as argument to have access to activity calls, which
```
johntyree: storageObj is not passed as an argument.

docs/store.js
```diff
@@ -1,26 +1,14 @@
 // Firebase storage manager.
 var firebaseCalendarStorage = function(params) {
```
johntyree: If this is a class created with `new` then it should start with a capital letter.

docs/store.js
```diff
@@ -424,7 +413,7 @@ var LocalCalendarStorage = function(params) {
             self.endWork();
         };
         
-        var randomNumber = Math.random() * 4000;
+        var randomNumber = Math.random() * 100;
```
johntyree: set the jitter value at construction time via `params`

tests/storetests.js
```diff
@@ -0,0 +1,235 @@
...
+        this.$fixture = $('#qunit-fixture');
+        this.$calendarDiv = this.$fixture.find('#calendarDiv');
+        
+        this.$fixture.append(
+            `<script>
+            $(document).ready( function() {
+            checkit = new CheckIt('localStorage');
+            });
+            </script>`);
+        this.checkit = checkit;
```
johntyree, line 40: This checkit is not the checkit you just created, is it? Are you trying to work around something here?

tests/tests.js
```diff
@@ -711,32 +711,42 @@ QUnit.test("validateForm test", function(assert) {
...
-  },
   beforeEach: function() {
     // prepare something before each test
+    this.$fixture = $('#qunit-fixture');
+    this.$calendarDiv = this.$fixture.find('#calendarDiv');
+    
+    this.$fixture.append(
+        `<script>
+        $(document).ready( function() {
+        checkit = new CheckIt('localStorage');
```
johntyree, line 722: What is this for?

johntyree (changes requested):

docs/checkit.js
andrayantelo: Fixed comment.

docs/store.js
andrayantelo: Fixed.

docs/store.js
andrayantelo: Did it, but why? Everywhere I use the same value for milliseconds.

tests/storetests.js
andrayantelo: I don't think it's the same checkit. I need to use checkit functions that's why I did line 40.

tests/tests.js
andrayantelo: For checkit's store initialization test.

docs/store.js
johntyree: As a general rule, hardcoding numbers into the middle of your code makes it difficult to update and understand later. By parameterizing it you can play with the value without rewriting everything all the time.

tests/storetests.js
johntyree: I'm still confused. Why do you need to create a checkit object on document ready as part of your test by inject a script tag into the page?

tests/tests.js
johntyree: Why can't you use the checkit that you already have?

docs/store.js
```diff
@@ -407,13 +407,13 @@ var LocalCalendarStorage = function(params) {
       
     };
     
-    var jitter = function(func, arg) {
+    var jitter = function(params) {
```
johntyree: So I guess this was my fault for not explaining what I meant further. I meant that the jitter size could be set by a value you pass to the storage at construction. If you needed to pass different values in different settings, then what you've done would make sense, but we don't really need that since all this is doing is simulating random delays.

johntyree (changes requested):

tests/tests.js
andrayantelo: Well, it must be the same because I don't initialize checkit anywhere else but in the script tag. I just wanted it to be as close to as it would work on the actual page as possible.

docs/store.js
andrayantelo: SIGH.

andrayantelo: Fixed.

johntyree (approved):

PR #170, opened 2017-03-14
Testing checkit functions that involve Storage (with localStorage)

andrayantelo:
Ready for review @johntyree I asked questions in comments

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -369,12 +369,12 @@ CheckIt.prototype.fillCalendar = function(calObj) {
 
     calObj.monthObjects.forEach (function(monthObj) {
         
-        var monthId = '#' + monthObj.monthId;
+        var $monthId = checkit.$calendarDiv.find('#' + monthObj.monthId).
         
-        checkit.$calendarDiv.find(monthId).find(".month-year").text(monthObj.monthName + " " + monthObj.monthYear);
+        $monthId.find(".month-year").text(monthObj.monthName + " " + monthObj.monthYear);
         
         // Go through each td and fill in correct day number
-        checkit.$calendarDiv.find(monthId).find($('.week')).find('td').each( function(indexOfTableTd) {
+        $monthId.find($('.week')).find('td').each( function(indexOfTableTd) {
             
             // The indexOfTableTd is where we are currently on the month table
             // which td are we in, from 0 to 41, because there are 6 rows
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -1,6 +1,6 @@
 
 // Initializes CheckIt
-function CheckIt(mode) {
+function CheckIt(mode, calendarDiv) {
     // Shortcuts to DOM elements.
     this.mode = mode;
     
@@ -31,8 +31,8 @@ function CheckIt(mode) {
     this.$startDate = $('#startDate');
     this.$endDate = $('#endDate');
     this.$buildCalendarForm = $('#collapseOne');
-    this.$calendarDiv = $('#calendarDiv');
     this.$loadingWheel = $('#loadingWheel');
+    this.$calendarDiv = calendarDiv;
     
     this.spinner = new Spinner();
     
@@ -157,14 +157,23 @@ CheckIt.prototype.fillDropdown = function($dropdown) {
         }.bind(this));
 };
 
+// TODO setCalendarDiv, clear page should only clear the div with the calendar in it
+// that div you pass as an argument, and don't pass the calendarDiv that lives on
+// checkit to any of these calendar methods
+
+CheckIt.prototype.setCalendarDiv = function($div) {
+    this.$calendarDiv = $div;
+};
+
 CheckIt.prototype.displayActiveCalendar = function($div) {
     // Display the active calendar if there is one.
       // Get the current active calendar from storage and display it.
     // If there is none, show build calendar menu.
     // $div is the div where you want to build the calendar
-   this.store.getActive()
+    console.log("displaying calendar on div:");
+    console.log($div);
+   return this.store.getActive()
        .then(function (activeCalendarId) {
-           
            this.store.loadById(activeCalendarId)
                .then(function (activeCalendarState) {
                    if (activeCalendarState !==  null) {
@@ -196,10 +205,10 @@ CheckIt.prototype.clearDropdown = function($dropdown) {
 CheckIt.prototype.initLocalStorage = function() {
     this.store = new LocalCalendarStorage({storeId: '', jitterTime: 100});
     // Fill the dropdown with user's saved calendar titles/
-    this.fillDropdown(this.$calendarDropdown);
+    //this.fillDropdown(this.$calendarDropdown);
     
     // Display the user's active calendar.
-    this.displayActiveCalendar(this.$calendarDiv);
+    //this.displayActiveCalendar(this.$calendarDiv);
     
     //Show the build Calendar form.
     this.$buildFormAccordion.removeAttr('hidden');
@@ -337,6 +346,8 @@ CheckIt.prototype.generateEmptyCalendar = function(calObj, $calendarDiv) {
     // display.
     var checkitApp = this;
     // Add the title of the calendar
+    console.log("Generating empty calendar for:");
+    console.log($calendarDiv);
     $calendarDiv.append('<div id="calendarTitleHeading"><h1 class="page-header\
                          text-center">'+calObj.state.title + '</h1></div>');
                         
@@ -359,7 +370,7 @@ CheckIt.prototype.generateEmptyCalendar = function(calObj, $calendarDiv) {
 
 CheckIt.prototype.fillCalendar = function(calObj) {
     // Fill an empty calendar with appropriate calendar data.
-    
+
     calObj.monthObjects.forEach (function(monthObj) {
         
         var $monthId = $('#'+ monthObj.monthId);
@@ -417,7 +428,8 @@ CheckIt.prototype.fillCalendar = function(calObj) {
 CheckIt.prototype.generateCheckmarks = function(calObj, $calendarDiv) {
     // Display saved checkmarks on calendar
     // div is which div do you want to look through for checkmarks
-    
+    console.log("generating checkmarks for:");
+    console.log($calendarDiv);
     if (calObj.state.checkedDays === undefined) {
         console.log("No days are checked");
         return;
@@ -436,7 +448,8 @@ CheckIt.prototype.generateCheckmarks = function(calObj, $calendarDiv) {
 
 CheckIt.prototype.removeEmptyWeeks = function(calObj, $calendarDiv) {
     // Remove empty weeks from the calendar
-    
+    console.log("removing empty weeks for");
+    console.log($calendarDiv);
     $calendarDiv.find('.month .week').each( function(index) {
         if ($(this).find('td > .nil').length === 7) {
             $(this).remove();
@@ -715,9 +728,10 @@ CheckIt.prototype.buildCalendar = function(calendarObject, $div) {
     //this function assumes the calendarObject already has it's
     //state updated with the correct information. 
     //$div is the div where you want to build the calendar
-
+    console.log("Running build Calendar for div");
+    console.log($div);
     this.generateEmptyCalendar(calendarObject, $div);
-    this.fillCalendar(calendarObject)
+    this.fillCalendar(calendarObject);
     this.attachCheckmarkClickHandler(calendarObject, calendarObject.monthObjects);
     this.generateCheckmarks(calendarObject, $div);
     this.removeEmptyWeeks(calendarObject, $div);
@@ -729,6 +743,8 @@ CheckIt.prototype.buildCalendar = function(calendarObject, $div) {
 CheckIt.prototype.displayCalendar = function(calendarObj, $div) {
     //load a state and build the calendar on the page
     // $div is the div where you want to build the calendar
+    console.log("Running displayCalendar in div:");
+    console.log($div);
     this.clearPage(this.$calendarDiv);
     this.buildCalendar(calendarObj, $div);
     this.store.setActiveById(calendarObj.state.uniqueId);
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -187,7 +187,7 @@ CheckIt.prototype.displayActiveCalendar = function() {
        }.bind(this))
        
        .catch(function (err) {
-           console.log("There is no current active calendar" + err);
+           console.log("There is no current active calendar " + err);
            this.showForm(this.$buildCalendarForm);
            return err;
        }.bind(this));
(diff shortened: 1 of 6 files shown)


docs/checkit.js
```diff
@@ -685,16 +687,22 @@ CheckIt.prototype.deleteCalendar = function() {
...
-                this.store.removeById(currentCalendarId);
-                // To delete a calendar you have to be looking at it and if you
-                // are looking at it that means it is the currentActiveCalendar
-                this.store.removeActive();
-                //console.log("clearing the page");
-                //clear the page
-                this.clearPage();
-                this.showForm(this.$buildCalendarForm); 
+                this.store.removeById(currentCalendarId)
+                    .then(function() {
```
andrayantelo: Added a .then here. Am I using it right? @johntyree

tests/tests.js
```diff
@@ -788,6 +788,9 @@ QUnit.test("displayActiveCalendar test", function(assert) {
     var displayActiveP = this.checkit.displayActiveCalendar();
     displayActiveP.then(function () {
         assert.equal($calendarDiv.find('h1').text(), 'hello');
+        console.log($calendarDiv.html());
```
johntyree: I assume you didn't mean to push these?

johntyree (changes requested):

docs/checkit.js
```diff
@@ -330,42 +338,43 @@ CheckIt.prototype.onAuthStateChanged = function(user) {
...
 
 CheckIt.prototype.fillCalendar = function(calObj) {
     // Fill an empty calendar with appropriate calendar data.
-    
+    var checkit = this;
+
     calObj.monthObjects.forEach (function(monthObj) {
         
-        var $monthId = $('#'+ monthObj.monthId);
+        var monthId = '#' + monthObj.monthId;
```
johntyree: You also could have changed this to be `var $monthId = checkit.$calendarDiv.find('#' + monthObj.monthId)`.

docs/checkit.js
```diff
@@ -623,29 +632,32 @@ CheckIt.prototype.createCalendar = function() {
     if (this.validateForm(start, end)) {
         
         //clear the previously displayed calendar
-        this.clearPage(this.$calendarDiv);
+        this.clearPage();
```
johntyree: This function is poorly named. It only clears `$calendarDiv`.

docs/checkit.js
```diff
@@ -623,29 +632,32 @@ CheckIt.prototype.createCalendar = function() {
...
 
         //make a calendar State
         var state = emptyCalendarState({startDate: start, endDate: end, calendarTitle: title});
         
         //make calendar object
 
         var calendar = new Calendar(state);
+        // TODO figure out if I should be returning initializeCalendar and buildCalendar promises
```
johntyree: I'd say so. Without doing it you can't write any code that depends on this being done. There's no real reason not to.

docs/checkit.js
```diff
@@ -673,18 +685,24 @@ CheckIt.prototype.deleteCalendar = function() {
...
                 //delete the calendar and remove its active calendar status
-                this.store.removeById(currentCalendarId);
-                // To delete a calendar you have to be looking at it and if you
-                // are looking at it that means it is the currentActiveCalendar
-                this.store.removeActive();
-                //console.log("clearing the page");
-                //clear the page
-                this.clearPage(this.$calendarDiv);
-                this.showForm(this.$buildCalendarForm); 
+                return this.store.removeById(currentCalendarId)
```
johntyree: You can avoid have deep nesting by letting the fact that handlers that return Promises will "pass them through". You don't have to, just pointing it out. Note that you do need to put the catch handlers in the right place to catch the error at the right "level."

```
var checkit = this;
return checkit.store.getActive().then(function(currentCalendarId) {
  checkit.removeBlahBlah();
  return checkit.store.removeById(currentCalendarId).catch(function() {
    console.log('couldn't remove id from storage');
  });
}).then(function() {
  checkit.clearPage();
  checkit.showForm(..);
  return checkit.store.removeActive().catch(function() {console.log("couldn't remove active from store or whatever")});
}).catch(function() {console.log('couldn't delete');})
```

docs/store.js
```diff
@@ -64,7 +64,7 @@ var FirebaseCalendarStorage = function(params) {
             .catch(function(err) {
                 self.endWork();
                 console.error(err);
-                return err;
+                return {};
```
johntyree: I'd add a comment here about why you're returning this.

docs/store.js
```diff
@@ -437,7 +436,8 @@ var LocalCalendarStorage = function(params) {
             })
             .catch( function(reason) {
                 console.log("getAllCalendarIds catch function running. " + reason);
-                return reason;
+                // returns a new promise which resolves with this value {}
```
johntyree: Comments are about explaining *why* you are doing something, not what it does. You can figure out this comment by reading the docs, but it's not clear why you did it.  E.g. "If there's no index of all calendar ids and their titles, make a new one and use that."

tests/tests.html
```diff
@@ -39,8 +39,8 @@
   <link rel="stylesheet" type="text/css" href="../docs/calendar.css">
 
   <!-- My tests -->
-  <script type="text/javascript" src="storetests.js"></script>
-  <!--<script type ="text/javascript" src="tests.js"></script>-->
+  <!--<script type="text/javascript" src="storetests.js"></script>-->
```
johntyree: What's going on here?

tests/tests.js
```diff
@@ -149,11 +149,27 @@ QUnit.module( "CheckIt Tests", {
 
 beforeEach: function() {
     // Prepare something once for all tests
```
johntyree: Better comment: `// Initialize a fresh `checkit` object and a dummy active calendar for every test`

tests/tests.js
```diff
@@ -720,86 +744,210 @@ QUnit.test("validateForm test", function(assert) {
...
     
     this.$fixture.append(
         `<script>
         $(document).ready( function() {
-        checkit = new CheckIt('localStorage');
+        checkit = new CheckIt('localStorage', $('#qunit-fixture #calendarDiv'));
         });
         </script>`);
+ 
     this.checkit = checkit;
```
johntyree: Why can't you create the checkit object here rather than injecting it into the page with a script tag?

tests/tests.js
```diff
@@ -720,86 +744,210 @@ QUnit.test("validateForm test", function(assert) {
...
+    //active calls > 0
+    assert.equal($('#loadingWheel').children().length, 0);
+    this.checkit.onActivityChanged(1, 'loadingWheel');
+    assert.equal($('#loadingWheel').children().length, 1);
+    
+    // active calls === 0
+    this.checkit.onActivityChanged(0, 'loadingWheel');
+    assert.equal($('#loadingWheel').children().length, 0);
+    
+    // it's possible for active calls to be less than zero, maybe I should have the loading wheel
```
johntyree: How do you manage to have a negative number of active calls? What does that even mean? Sounds like a bug.

tests/tests.js
```diff
@@ -720,86 +744,210 @@ QUnit.test("validateForm test", function(assert) {
...
+        assert.equal(noActive, null);
+        assert.deepEqual(state, null);
+        assert.equal(ids['123'], null);
+        done();
+    });
 });
 
 QUnit.test("buildCalendar test", function(assert) {
+    assert.expect(0);
     // Probably not needed because each function inside of buildCalendar
```
johntyree, line 871: You'll laugh when you have to delete this comment somday :)

tests/tests.js
```diff
@@ -720,86 +744,210 @@ QUnit.test("validateForm test", function(assert) {
...
-  after: function() {
-    // clean up once after all tests are done
   }
 });
 
 QUnit.test("initFirebase test", function(assert) {
     assert.expect(0);
+    //console.log(this.checkit.mode);
+    
+    // Can only do these tests if I include the firebase key in my html.
```
johntyree, line 933: Good question. What do the firebase docs say about writing tests that need to use firebase?

tests/tests.js
```diff
@@ -821,16 +969,15 @@ QUnit.test("onAuthStateChanged test", function(assert) {
...
   afterEach: function() {
     // clean up after each test
-  },
-  after: function() {
-    // clean up once after all tests are done
   }
 });
+
+
+// Have to have the same html in test html as my app html so that I can
```
johntyree, line 965: If you want to test that your buttons and stuff work, then yes probably. I'd caution against making sure that the correct functions are called though. Just make sure that the correct thing has happened. i.e. check that a checkmark has appeared, not that "create_checkmark" was called or whatever it is.

johntyree (changes requested):

johntyree:
It's much easier for me to notice your questions if you review your own code and leave them as github comments.

johntyree:
I.e. after you push, go up to the "files changed" tab and leave comments on the lines you have questions about.

johntyree:
Or alternatively, you could just leave an empty comment to draw attention to the questions you've written directly into the code.

docs/checkit.js
andrayantelo: Yes, better idea.

docs/checkit.js
andrayantelo: Changed to clearCalendarPage.

docs/checkit.js
andrayantelo: Okay.

tests/tests.html
andrayantelo: I just wanted to run the one test file and not the other.

tests/tests.js
andrayantelo: I wanted to replicate the way things work in the code as much as possible. Should I not do that?

tests/tests.js
andrayantelo: Actually, I'm not sure if this is possible. I think several weeks ago I saw it happen. lol

andrayantelo: months ago maybe even.

tests/tests.js
andrayantelo: So should I make some test out of this?

andrayantelo: I think I was also really sick and tired of writing tests at that point.

tests/tests.js
andrayantelo: I didn't find anything useful when I looked. Some people tried to make a mock firebase thing that you could write tests against but it's not reliable.

tests/tests.js
andrayantelo: Yeah. That's what I meant when I wrote that. Ugh more tests.

andrayantelo: I'll probably start a new branch for these new tests.

tests/tests.js
johntyree: That's a good goal, but I'm not sure you *want* to have a checkit object that requires a fully loaded page before it can be created. Does it do page manipulation while it's being created?

tests/tests.js
johntyree: This is the kind of error that you don't really want to "fix" because it means your UI is going to be doing wrong/bad things. For example, if the count is negative and I start doing work, then the spinner won't appear because the count is not positive.

I'd assert that active calls is >= 0 whenever you decrement it.

tests/tests.js
johntyree: Nope. It's just one of those "famous last words" kind of comments.

tests/tests.js
johntyree: Maybe let's kick this down the road a bit farther then. If we've checked that the app behaves correctly assuming storage is working, then we only have to assume firebase works, which it probably does. Eventually we'll want tests that include actual reading and writing to storage to protect ourselves from e.g. breaking things due to badly configured permissions.

File a bug to investigate integration testing with firebase. I know they provide some kind of test environment, but it's probably not trivial to set up.

tests/tests.js
johntyree: Testing is good, but you should pace yourself. Tests have to be maintained too. If you are still making UI changes then you're going to be breaking a lot of UI tests.

UI testing is not that common in projects like this because the effort required is pretty significant, but the benefit is marginal. If the UI breaks, it's usually pretty clear that things aren't working.

johntyree (approved):

johntyree (changes requested):
Oops.

tests/tests.js
```diff
@@ -151,25 +151,11 @@ beforeEach: function() {
...
     this.$calendarDiv = this.$fixture.find('#calendarDiv');
-    this.$fixture.append(
-        `<script>
-        $(document).ready( function() {
-        checkit = new CheckIt('localStorage', $('#qunit-fixture #calendarDiv'));
-        });
-        </script>`);
  
-    this.checkit = checkit;
+    this.checkit = new CheckIt('localStorage', $('#qunit-fixture #calendarDiv'));
```
johntyree: On that note, it's probably a good idea to log a warning during initialization if checkit is initialized with a selector that doesn't select anything.

docs/store.js
```diff
@@ -395,11 +395,13 @@ var LocalCalendarStorage = function(params) {
     
     self.endWork = function() {
         // Will decrement the counter and maybe fire an event.
-        
-        
+
         self.activeCalls -= 1;
+        
         if (self.activeCalls < 0) {
             console.error("No work has been started");
```
johntyree: This error message could be better.  e.g. "Number of active calls is negative." 

The entire point is that you don't really know if there's work happening or not because clearly the counting system is broken if you're getting negative numbers.

johntyree (changes requested):

docs/checkit.js
```diff
@@ -32,7 +32,13 @@ function CheckIt(mode, calendarDiv) {
     this.$endDate = $('#endDate');
     this.$buildCalendarForm = $('#collapseOne');
     this.$loadingWheel = $('#loadingWheel');
-    this.$calendarDiv = calendarDiv;
+    
+    if (calendarDiv.length === 0) {
+        console.error("Selector given for calendar div does not select anything.");
+    }
+    else {
+        this.$calendarDiv = calendarDiv;
```
johntyree: I would always do this. Either you do what the client asked you to do (assign this value) or you refuse to do anything at all. The worst thing to do is to magically change your behavior based on a best guess about what should happen. You've warned them that what is about to happens seems stupid. That's enough.

johntyree: You could treat the case where `calendarDiv` itself is undefined (i.e. the caller didn't pass a second argument) as meaning "I don't want to initialize this yet, i'll do it later. That could be fine.

johntyree (changes requested):

docs/checkit.js
```diff
@@ -36,9 +36,8 @@ function CheckIt(mode, calendarDiv) {
     if (calendarDiv.length === 0) {
```
johntyree: This will blow up if `calendarDiv` is undefined...

johntyree (changes requested):
Guess you needed that test case after all...

johntyree (approved):
:1st_place_medal:

johntyree:
:tada:

PR #185, opened 2017-04-18
Github fork me image included

andrayantelo:
Resized for mobile devices.

### DIFF
--- a/docs/calendar.css
+++ b/docs/calendar.css
@@ -108,6 +108,19 @@ html {
   padding-top:10px;
 }
 
+/* Github fork me image CSS
+ * **********************/
+ 
+#forkMe {
+  position: absolute;
+  top: 50px;
+  right: 0;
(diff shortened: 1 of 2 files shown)


docs/calendar.css
```diff
@@ -361,6 +374,11 @@ corresponding value from the list below*/
     float: right;
   }
   
+  #forkMe {
```
johntyree, line 377: Why is this repeated?

johntyree (changes requested):

docs/calendar.css
andrayantelo: It gets resized for tablet and larger screen sizes.

johntyree: Oh I see. The diff on GitHub obscures the fact that this is inside of a `@media` query.

johntyree (approved):

PR #186, opened 2017-04-18
Functions that return calendar facts

### LINKED ISSUE #190: Make a calendarAnalyzer object
The new object CalendarAnalyzer's functions: 
getNumberOfChecked - Returns the number of checked days in a calendar
getTotalCalendarDays - Returns the total number of active days in calendar
getNumberOfUnchecked - Returns number of unchecked days in calendar
getCheckedDaysStreak - Returns the longest streak of checked days
getUncheckedDaysStreak - Returns the longest streak of unchecked days
getTotalCalendarWeeks - Returns total number of active weeks in calendar
getNumOfDaysLeft - Returns total number of active days left in calendar
starting from current day ending at end of calendar

Tests Done:
- [ ] getNumberOfChecked
- [ ] getNumberOfUnchecked
- [ ] getCheckedDaysStreak
- [ ] getUncheckedDaysStreak
- [ ] getTotalCalendarDays + getTotalCalendarWeeks
- [ ] getNumOfDaysLeft

andrayantelo:
Such as:
- Number of days checked so far
- Number of unchecked days in the calendar
- Longest streak of checked days
- Longest streak of unchecked days
- Total number of calendar days
- Total number of calendar weeks
- How many days are left in the calendar

Resolves #190

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -163,16 +163,12 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
     "use strict";
     
     var currentStreak = 1,
-        checkedDaysArray = [],
+        checkedDaysArray = (Object.keys(this.calState.checkedDays)).sort(),
         bestStreak = 0,
         i,
         currentDay,
         previousDay,
         dayDiff;
-    
-    // Place checkedDays object's keys into array and sort the array.
-    checkedDaysArray = Object.keys(this.calState.checkedDays);
-    checkedDaysArray.sort();
     
     previousDay = moment(checkedDaysArray[0], "YYYYMMDD");
     
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -162,36 +162,42 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
     // Returns the longest streak of checked days
     "use strict";
     
-    var currentStreak = 0,
+    var currentStreak = 1,
         checkedDaysArray = [],
-        lastDate,
         bestStreak = 0,
         i,
         currentDay,
         previousDay,
         dayDiff;
     
-    // place checkedDays object's keys into array and sort the array.
+    // Place checkedDays object's keys into array and sort the array.
     checkedDaysArray = Object.keys(this.calState.checkedDays);
     checkedDaysArray.sort();
-
-    lastDate = checkedDaysArray[0];
+    
+    previousDay = moment(checkedDaysArray[0], "YYYYMMDD");
     
     for (i = 1; i < checkedDaysArray.length; i++) {
-        // if the next element is the day after lastDate, add 1 to currentStreak
-        currentDay = moment(checkedDaysArray[i], "YYYYMMDD");
-        previousDay = moment(checkedDaysArray[i - 1], "YYYYMMDD");
         
+        currentDay = moment(checkedDaysArray[i], "YYYYMMDD");
+        
+        // Calculate how many days are between previous day
+        // and current day
         dayDiff = currentDay.diff(previousDay, 'days');
-
-        currentStreak++;
+        
+        // If they are not consecutive days 
         if (dayDiff !== 1) {
-            if (bestStreak < currentStreak) {
-                bestStreak = currentStreak;
-            }
+            // Find bestStreak
+            bestStreak = Math.max(bestStreak, currentStreak);
             // reset currentStreak
-            currentStreak = 0;
+            currentStreak = 1;
+        } else {
+            // Increment currentStreak
+            currentStreak++;
+            // Find bestStreak
+            bestStreak = Math.max(bestStreak, currentStreak);
         }
+        // Update previousDay
+        previousDay = currentDay;
     }
 
     return bestStreak;
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -202,10 +202,34 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
 CalendarAnalyzer.prototype.getUncheckedDaysStreak = function () {
     // Returns the longest streak of unchecked days
     "use strict";
-    // TODO
-    // Go through all the dates between startDate and endDate, if they
-    // are not in checkedDays add them to an uncheckedDays array
-    // then find the bestStreak in that uncheckedDays array. 
+
+    // The largest gap between any two checked days (minus one)
+    var checkedDaysArray = (Object.keys(this.calState.checkedDays)).sort(),
+        currentGap = 0,
+        longestGap = 0,
+        i,
+        currentDay,
+        previousDay,
+        dayDiff;
+    
+    previousDay = moment(checkedDaysArray[0], "YYYYMMDD");
+    
+    for (i = 1; i < checkedDaysArray.length; i++) {
+        currentDay = moment(checkedDaysArray[i], "YYYYMMDD");
+        
+        dayDiff = currentDay.diff(previousDay, 'days');
+
+        if (dayDiff === 1) {
+            longestGap = Math.max(longestGap, currentGap);
+            currentGap = 0;
+        } else {
+            currentGap = dayDiff;
+            longestGap = Math.max(longestGap, currentGap);
+        }
+        previousDay = currentDay;
+    }
+
+    return longestGap;
 
 };
 
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -782,6 +782,36 @@ CheckIt.prototype.findCurrentDay = function() {
     return todayId;
 };
 
+CheckIt.prototype.getNumberOfCheckedDays = function() {
+    // Returns the number of checked days in a calendar
+};
+
+CheckIt.prototype.getNumberOfUncheckedDays = function() {
+    // Returns number of unchecked days in a calendar
+};
+
+CheckIt.prototype.getCheckedDaysStreak = function() {
+    // Returns the longest streak of checked days
+};
+
+CheckIt.prototype.getUncheckedDaysStreak = function() {
+    // Returns the longest streak of unchecked days
+};
+
+CheckIt.prototype.getTotalCalendarDays = function() {
+    // Returns the total number of active days in the calendar
+};
+
+CheckIt.prototype.getTotalCalendarWeeks = function() {
+    // Returns the total number of active weeks in the calendar
+};
+
+CheckIt.prototype.getNumOfDaysLeft = function() {
+    // Returns the total number of active days left in a calendar starting from current
+    // day to end.
+};
+
+
 
 //CODE FOR MONTH AND CALENDAR OBJECTS
 var Month = function(dateString) {
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -870,15 +870,20 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function() {
     return arrayOfStreakArrays[largestArrayIndex].length;
 };
 
-CalendarAnalyzer.prototype.getUncheckedDaysStreak = function(calObj) {
+CalendarAnalyzer.prototype.getUncheckedDaysStreak = function() {
     // Returns the longest streak of unchecked days
 };
 
-CalendarAnalyzer.prototype.getTotalCalendarWeeks = function(calObj) {
+CalendarAnalyzer.prototype.getTotalCalendarWeeks = function() {
     // Returns the total number of active weeks in the calendar
+    
+    var endDate = moment(this.calState.endDateString, "YYYYMMDD");
+    var startDate = moment(this.calState.startDateString, "YYYYMMDD");
+    var totalWeeks = endDate.diff(startDate, 'weeks');
+    return totalWeeks;
 };
 
-CalendarAnalyzer.prototype.getNumOfDaysLeft = function(calObj) {
+CalendarAnalyzer.prototype.getNumOfDaysLeft = function() {
     // Returns the total number of active days left in a calendar starting from current
     // day to end.
 };
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -782,6 +782,9 @@ CheckIt.prototype.findCurrentDay = function() {
     return todayId;
 };
 
+
+// Should these be CheckIt functions? or Calendar. TODO
+
 CheckIt.prototype.getNumberOfChecked = function(calObj) {
     // Returns the number of checked days in a calendar
     var size = 0, key;
@@ -793,8 +796,9 @@ CheckIt.prototype.getNumberOfChecked = function(calObj) {
     return size;
 };
 
-CheckIt.prototype.getNumberOfUncheckedDays = function(calObj) {
+CheckIt.prototype.getNumberOfUnchecked = function(calObj) {
     // Returns number of unchecked days in a calendar
+
 };
 
 CheckIt.prototype.getCheckedDaysStreak = function(calObj) {
@@ -807,6 +811,9 @@ CheckIt.prototype.getUncheckedDaysStreak = function(calObj) {
 
 CheckIt.prototype.getTotalCalendarDays = function(calObj) {
     // Returns the total number of active days in the calendar
+    
+    var totalDays = calObj.endDate.diff(calObj.startDate, 'days');
+    return totalDays;
 };
 
 CheckIt.prototype.getTotalCalendarWeeks = function(calObj) {
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -210,19 +210,13 @@ CalendarAnalyzer.prototype.getUncheckedDaysStreak = function () {
         i,
         currentDay,
         previousDay,
-        dayDiff;
+        dayDiff,
+        lastDay;
     
-    // what if the first checked day doesn't occur until we are
-    // well into the calendar? It's possible that the time before
-    // the first checked day is the longest streak of unchecked days
-    // this could be true for the time that comes after the last checked
-    // day. Therefore, we need to begin checking with the start day and
-    // the first checked day, and 
-    // end with comparing the last checked day with the last day in
-    // in the calendar TODO
-    previousDay = moment(checkedDaysArray[0], "YYYYMMDD");
+    previousDay = moment(this.calState.startDateString, "YYYYMMDD");
+    lastDay = moment(this.calState.endDateString, "YYYYMMDD");
     
-    for (i = 1; i < checkedDaysArray.length; i++) {
+    for (i = 0; i < checkedDaysArray.length; i++) {
         currentDay = moment(checkedDaysArray[i], "YYYYMMDD");
         
         dayDiff = currentDay.diff(previousDay, 'days');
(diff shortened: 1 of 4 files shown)


docs/checkit.js
```diff
@@ -783,6 +783,89 @@ CheckIt.prototype.findCurrentDay = function() {
...
+
+CheckIt.prototype.getNumberOfChecked = function() {
+    // Returns the number of checked days in a calendar
+
+    var checkit = this;
+
+    return checkit.store.getActive().then(function(activeCalendarId) {
+        console.log("found this activeCalendarId: " + activeCalendarId);
+        return checkit.store.loadById(activeCalendarId).catch(function(err) {
+            console.log("Couldn't find active calendar specifically in getNumberOfChecked");
```
johntyree: If these are errors, use console.error

docs/checkit.js
```diff
@@ -783,6 +783,89 @@ CheckIt.prototype.findCurrentDay = function() {
...
+            console.log("There is no current active calendar " + err);
+            checkit.store.removeActive();
+            return err;
```
johntyree: Returning a value here is going to "fix" the error and cause `.then()` to be called with `err` as the argument, isn't it?

johntyree (changes requested):

docs/checkit.js
andrayantelo: They aren't really errors. I was just using console.log to figure out what was happening when I was running the function and what wasn't happening. But I will change to console.error since those pop up more anyway.

docs/checkit.js
andrayantelo: Yeah, I will change that. My  main question with these functions are, should they be checkit functions? Should I be trying to get the active calendar to get the stats on it with these functions (I think yes)?

johntyree: It's up to you really. I could easily see an argument for making the analysis of calendars an entirely separate object.

Maybe a CalendarAnalyzer object that is given a calendar state and then can produce facts about it as needed. This has the additional benefit of letting you show information about multiple calendars by just creating more than one CalendarAnalyzer, etc.

docs/checkit.js
```diff
@@ -793,38 +792,40 @@ CheckIt.prototype.getNumberOfChecked = function() {
...
+                return checkit.displayCalendar(calendar);
             }
-            
         })
         .catch(function(err) {
-            console.log("getNumberOfChecked");
-            console.log("Could not get number of checked days" + err);
+            console.log("Could not display active calendar" + err);
+            checkit.showForm(checkit.$buildCalendarForm);
             return err;
```
johntyree: If the function is supposed to return a number and you can't then that's an error.

docs/checkit.js
```diff
@@ -793,38 +792,40 @@ CheckIt.prototype.getNumberOfChecked = function() {
...
-                console.log(size);
-                return size;
+                return checkit.displayCalendar(calendar);
             }
-            
         })
         .catch(function(err) {
-            console.log("getNumberOfChecked");
-            console.log("Could not get number of checked days" + err);
+            console.log("Could not display active calendar" + err);
```
johntyree: Why do you need to display an a calendar in order to count the number of checked days it has?

johntyree (changes requested):

docs/checkit.js
andrayantelo: lol.. no.. this is just I copy pasted form displayActiveCalendar function. This is still WIP.

docs/checkit.js
```diff
@@ -782,59 +782,30 @@ CheckIt.prototype.findCurrentDay = function() {
...
-        });
-            * 
-            * 
-            */
-            
-            
-            
-        });
+CalendarAnalyzer.prototype.getNumberOfChecked = function() {
+    // Returns the number of checked days in a calendar
```
johntyree: `return Object.keys(this.calState.checkDays).length`

docs/checkit.js
```diff
@@ -782,59 +782,30 @@ CheckIt.prototype.findCurrentDay = function() {
...
     
+    var size = 0, key;
+    for (key in this.calState.checkedDays) {
```
johntyree: I think you can declare variables in the loop directly, although it doesn't really change anything semantically.

`for (var key in this...)`

tests/tests.js
```diff
@@ -828,7 +828,7 @@ QUnit.test("onActivityChanged test", function(assert) {
     
 });
 
-QUnit.test("deleteCalendar test", function(assert) {
+QUnit.test("deleteCalendar test", function(assert) {
```
johntyree: Trailing whitespace?

johntyree (changes requested):

docs/checkit.js
```diff
@@ -811,26 +811,58 @@ CalendarAnalyzer.prototype.getNumberOfUnchecked = function(calObj) {
...
+            arrayOfStreakArrays.push(currentStreak);
+        }
+        
+    }
+    
+    // Find the array with the most elements in the arrayOfStreakArrays
+    var largestArrayIndex = arrayOfStreakArrays.reduce(function(maxI, el, i, arr) {
+        return el.length > arr[maxI].length ? i : maxI;}, 0);
+   
+    return arrayOfStreakArrays[largestArrayIndex].length;
```
andrayantelo: Please critique. @johntyree

docs/checkit.js
```diff
@@ -811,6 +811,11 @@ CalendarAnalyzer.prototype.getNumberOfUnchecked = function(calObj) {
 CalendarAnalyzer.prototype.getCheckedDaysStreak = function() {
     // Returns the longest streak of checked days
     
+    // if checkedDays is an empty object, return 0;
+    if (jQuery.isEmptyObject(this.calState.checkedDays)) {
+        return 0;
+    }
+
```
andrayantelo: added this to the function, in case this was one of your critique points @johntyree

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
 
 };
 
+// calendarAnalyzer takes a calendar object and can analyze facts about it.
+
+var CalendarAnalyzer = function (calendarState) {
+    "use strict";
```
johntyree, line 136: Can't you just put "use strict" at the top of the file?

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    var endDate = moment(this.calState.endDateString, "YYYYMMDD"),
+        startDate = moment(this.calState.startDateString, "YYYYMMDD"),
+        totalDays = endDate.diff(startDate, 'days');
+    return totalDays;
+};
+
+CalendarAnalyzer.prototype.getNumberOfUnchecked = function (calObj) {
+    // Returns number of unchecked days in a calendar
+    "use strict";
+    return this.getTotalCalendarDays(calObj) - this.getNumberOfChecked(calObj);
```
johntyree: This is good.

docs/checkit.js
johntyree: This is the second best solution. The best solution is to write the function in a way that "just works" with empty input.

For example if you imagine taking a sum of integers in a list. You could write a special case for the empty list and return 0 or you could start with 0, add all the elements of the list to it, then return it. In the second case, the empty list is handled because "add all the elements" will do nothing.

This doesn't mean that you can do that here. I just mean that in general this is the second best way to deal with this problem. It's certainly better than not dealing with it at all :)

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+        lastDate,
+        checkedDaysArrayLength,
+        i,
+        currentDay,
+        previousDay,
+        dayDiff,
+        largestArrayIndex;
+    
+    // place checkedDays object's keys into array and sort the array.
+    for (key in this.calState.checkedDays) {
```
johntyree: You can avoid having to check `hasOwnProperty` if you use `Object.keys()`

```
for (key in Object.keys(this.calState.checkedDays)) {
   ....
```

But even more than that, the [`Array.prototype.push`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) function you're using takes multiple elements at once. No loop needed. We use the [`Function.prototype.apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) method on the `push` function to expand our array of keys into individual arguments.

`checkedDaysArray.push.apply(checkedDaysArray, Object.keys(this.calState.checkedDays))`

But really what you're doing here is initializing checkDaysArray to whatever the keys are, right? So we can just use the keys array directly and be done with it.

Initialize it with `checkedDaysArray = Object.keys(this.calstate.checkedDays)` instead of `[]`

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+        if (this.calState.checkedDays.hasOwnProperty(key)) {
+            checkedDaysArray.push(key);
+        }
+    }
+    checkedDaysArray.sort();
+
+    checkedDaysArrayLength = checkedDaysArray.length;
```
johntyree: Why have this? In the best case it's the same as `checkedDaysArray.length`. In the worst case, the array changes and it's wrong.

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    
+    // for each checked day
+    for (i = 0; i < checkedDaysArrayLength; i++) {
+        
+        if (i === 0) {
```
johntyree: You know when `i === 0`.

Move this block to before the loop and start the loop at `i = 1`

Also it's not typical to declare `i` at the top of the function when it's just an iteration variable like this. You can declare it here. `for (var i = 1....)`  Does the linter want this?

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+};
+
+CalendarAnalyzer.prototype.getNumberOfUnchecked = function (calObj) {
+    // Returns number of unchecked days in a calendar
+    "use strict";
+    return this.getTotalCalendarDays(calObj) - this.getNumberOfChecked(calObj);
+};
+
+CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
+    // Returns the longest streak of checked days
```
johntyree: I think this function is too long and complicated. Figure out how to narrow it down and only do the absolute minimum amount of work to get the answer you need. For example [spoilers]: http://codepad.org/idgj5oyR

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    
+    // Find the array with the most elements in the arrayOfStreakArrays
+    largestArrayIndex = arrayOfStreakArrays.reduce(function (maxI, el, i, arr) {
+        return el.length > arr[maxI].length ? i : maxI;
+    }, 0);
+   
+    return arrayOfStreakArrays[largestArrayIndex].length;
+};
+
+CalendarAnalyzer.prototype.getUncheckedDaysStreak = function () {
```
johntyree: This one should be even simpler than the other one :+1:

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    // Returns the longest streak of unchecked days
+    "use strict";
+    // Probably don't need this right?
+};
+
+CalendarAnalyzer.prototype.getTotalCalendarWeeks = function () {
+    // Returns the total number of active weeks in the calendar
```
johntyree: What is "active" here?

docs/checkit.js
```diff
@@ -130,6 +130,128 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    "use strict";
+    var endDate = moment(this.calState.endDateString, "YYYYMMDD"),
+        startDate = moment(this.calState.startDateString, "YYYYMMDD"),
+        totalWeeks = endDate.diff(startDate, 'weeks');
+    return totalWeeks;
+};
+
+CalendarAnalyzer.prototype.getNumOfDaysLeft = function () {
+    // Returns the total number of active days left in a calendar starting from current
```
johntyree: Again I don't know what "active" means in this context. Days with checks? Days that could be checked? What's an "inactive" day?

johntyree: Oh.. maybe like a countdown until the calendar is "done?"

johntyree (changes requested):

docs/checkit.js
andrayantelo: jslint wants me to do it in every function. Should I not do that?

docs/checkit.js
andrayantelo: the linter wants that yes.

andrayantelo: It wants you to declare it with the other variables.

docs/checkit.js
andrayantelo: active days... like... you could start your calendar on January 5th and end it on January 31st, then Jan 1st - 4th are inactive days, and the rest of the days in January are active.

docs/checkit.js
andrayantelo: I was thinking i didn't want to have to run .length every time in the for loop. Does that not matter?

docs/checkit.js
andrayantelo: But do I need it? lol

andrayantelo: How is this simpler without using the page's HTML? I mean, calendarState does not have an uncheckedDays object. 
You have the startDate and the endDate and the days that are checked. So then you could maybe go through all the dates between startDate and endDate and check if they are in checkedDays and if they are not put them in an unCheckedDays array, and then look for the streak. That doesn't sound simpler lol

docs/checkit.js
johntyree: It's fine to do, I just thought you could get the whole file at once. Seems kind of annoying to have this scattered everywhere. JS is noisy enough.

docs/checkit.js
```diff
@@ -152,85 +152,59 @@ CalendarAnalyzer.prototype.getTotalCalendarDays = function () {
...
-            currentStreak = [];
-            currentStreak.push(checkedDaysArray[i]);
-        }
-        
-        // if on last element, push the currentStreak into the array of streaks
-        if (i === checkedDaysArrayLength - 1) {
-            arrayOfStreakArrays.push(currentStreak);
+        currentStreak++;
+        if (dayDiff !== 1) {
+            if (bestStreak < currentStreak) {
```
johntyree: `bestStreak = Math.max(bestStreak, currentStreak);`

docs/checkit.js
johntyree: `[1,2,3,4,7,8,15,16]`

What's the largest gap? How did you come up with it?

docs/checkit.js
```diff
@@ -152,85 +152,59 @@ CalendarAnalyzer.prototype.getTotalCalendarDays = function () {
...
     }
-    
-    // Find the array with the most elements in the arrayOfStreakArrays
-    largestArrayIndex = arrayOfStreakArrays.reduce(function (maxI, el, i, arr) {
-        return el.length > arr[maxI].length ? i : maxI;
-    }, 0);
-   
-    return arrayOfStreakArrays[largestArrayIndex].length;
+
+    return bestStreak;
```
johntyree: This will return zero until there is are at least two streaks, otherwise you'll never have `daydiff !== 1`.

johntyree (changes requested):

docs/checkit.js
andrayantelo: Of course!

docs/checkit.js
andrayantelo: Updated.

johntyree:
If you have a branch that contains all the changes from another branch, you should mention that in the description so that it's clear what's going on. It's confusing to see the same changes scattered around in different branches. I don't know which parts I need to review and which parts are just there because you didn't start the branch from master.

docs/checkit.js
```diff
@@ -162,36 +162,42 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
...
-
-        currentStreak++;
+        
+        // If they are not consecutive days 
         if (dayDiff !== 1) {
-            if (bestStreak < currentStreak) {
-                bestStreak = currentStreak;
-            }
+            // Find bestStreak
+            bestStreak = Math.max(bestStreak, currentStreak);
```
johntyree: You do this in both branches of the `if` statement. Move it after.

docs/checkit.js
```diff
@@ -162,36 +162,42 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
...
         dayDiff;
     
-    // place checkedDays object's keys into array and sort the array.
+    // Place checkedDays object's keys into array and sort the array.
     checkedDaysArray = Object.keys(this.calState.checkedDays);
     checkedDaysArray.sort();
-
-    lastDate = checkedDaysArray[0];
+    
+    previousDay = moment(checkedDaysArray[0], "YYYYMMDD");
```
johntyree: This doesn't throw an error when checkedDaysArray is empty?

docs/checkit.js
```diff
@@ -162,36 +162,42 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
...
     
     for (i = 1; i < checkedDaysArray.length; i++) {
```
johntyree: When there's only one element we don't enter this loop, right? How does `bestStreak` end up as `1`? When does it get incremented? I see you have a test for it, but I don't understand why it passes.

johntyree (changes requested):

docs/checkit.js
andrayantelo: currentStreak = 1 when I define it, on line 165.

docs/checkit.js
andrayantelo: But for one of them I need to reset currentStreak after I find bestStreak and for the other I need to increment currentStreak before finding bestStreak.

docs/checkit.js
andrayantelo: moment(undefined) is equivalent to moment(), which assumes the initial state to be the current date/time.

andrayantelo:
@johntyree 

> If you have a branch that contains all the changes from another branch, you should mention that in the description so that it's clear what's going on. It's confusing to see the same changes scattered around in different branches. I don't know which parts I need to review and which parts are just there because you didn't start the branch from master.

What are you talking about?

docs/checkit.js
andrayantelo: Although, when I test it in the console, I just get a moment object with the date being January 1st, and the year is just 0.

docs/checkit.js
```diff
@@ -202,10 +202,34 @@ CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
...
+            longestGap = Math.max(longestGap, currentGap);
+            currentGap = 0;
+        } else {
+            currentGap = dayDiff;
+            longestGap = Math.max(longestGap, currentGap);
+        }
+        previousDay = currentDay;
+    }
+
+    return longestGap;
```

> The largest gap between any two checked days (minus one)

Is it minus one or minus two? Since, previousDay and currentDay would both be checked days.

PR #188, opened 2017-04-20
Changing indentation for chained promises and fixing jslint errors

andrayantelo:
resolves #188

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -175,26 +175,67 @@ CheckIt.prototype.displayActiveCalendar = function() {
     
     var checkit = this;
 
-    return checkit.store.getActive().then(function(activeCalendarId) {
-        return checkit.store.loadById(activeCalendarId).catch(function(err) {
-            console.log("There is no current active calendar " + err);
-            checkit.store.removeActive();
-            checkit.showForm(checkit.$buildCalendarForm);
-            return err;
-            });
+    // if the first promise doesn't work, then just go to catch, don't
+    // try to 
+    return checkit.store.getActive()
+        .then(function(activeCalendarId) {
+            return checkit.store.loadById(activeCalendarId)
+                .catch(function(err) {
+                    console.log("There is no current active calendar " + err);
+                    checkit.store.removeActive();
+                    checkit.showForm(checkit.$buildCalendarForm);
+                    return err;
+                });
         })
-        .then(function(activeCalendarState) {
-            if (activeCalendarState !== null) {
-                var state = activeCalendarState;
-                var calendar = new Calendar(state);
-                return checkit.displayCalendar(calendar);
-            }
-        })
+                .then(function(activeCalendarState) {
+                    if (activeCalendarState !== null) {
+                        var state = activeCalendarState;
+                        var calendar = new Calendar(state);
+                        return checkit.displayCalendar(calendar);
+                    }
+                })
         .catch(function(err) {
             console.log("Could not display active calendar" + err);
             checkit.showForm(checkit.$buildCalendarForm);
             return err;
         });
+        
+        
+        /*
+         * 
+        return checkit.store.getActive()
+        .catch(function(err) {
+            console.error("Couldn't get active calendar id: " + err);
+        })
+        .then(function(currentCalendarId) {
+            // remove calendar from dropdown
+            checkit.removeFromCalendarDropdown(currentCalendarId,
+                                               checkit.$calendarDropdown);
+
+            // remove calendar state from storage
+            return checkit.store.removeById(currentCalendarId)
+                .catch(function(err) {
+                    console.error("Unable to remove calendar from storage.");
+                    // TODO: yes this is an error, but what do we want to do now?
+                    // Do we want to plow onward and clear the div or is this worth
+                    // stopping the entire app over?
+                    return Promise.reject(err);
+                })
+                .then(function() {
+                    checkit.clearCalendarDiv();
+                    checkit.showForm(checkit.$buildCalendarForm);
+                    return checkit.store.removeActive()
+                        .catch(function(err) {
+                            console.error("Unable to remove active status from calendar");
+                            return Promise.reject(err);
+                        });
+                });
+        });
+        * 
+        * 
+        */
+        
+        
 }
 
 CheckIt.prototype.clearDropdown = function($dropdown) {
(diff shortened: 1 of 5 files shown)


docs/checkit.js
```diff
@@ -685,30 +685,43 @@ CheckIt.prototype.deleteCalendar = function() {
...
             // remove calendar state from storage
-            return checkit.store.removeById(currentCalendarId).catch(function(err) {
-                console.log("Unable to remove calendar from storage.");
-                return err;
+            return checkit.store.removeById(currentCalendarId)
+                .catch(function(err) {
+                    console.error("Unable to remove calendar from storage.");
+                    // TODO: yes this is an error, but what do we want to do now?
+                    // Do we want to plow onward and clear the div or is this worth
+                    // stopping the entire app over?
```
andrayantelo: Well, if a person is looking at their calendar, and then they click on delete, but for some reason the calendar isn't even in the database... well they wanted to delete it anyway, right? What else would we do here?

andrayantelo: this .catch returns a rejected promise, which means the following
.then function does not get run.
 if we get this error, we could potentially handle it in the
 following .then function but we'd need to return a normal value
or a resolved promise...
                    
  so, I guess, if this is run because we are unable to remove a calendar
 from storage, then... we would still want the .then to run to remove 
the html of the calendar right?
 If we get to this catch function then there is something really wrong 
with the code, maybe it's better that the website just not work instead of
plowing on so that we know that something is seriously wrong? the way it's written now it's not plowing on

docs/store.js
```diff
@@ -276,6 +276,7 @@ var FirebaseCalendarStorage = function(params) {
             .catch(function(err) {
                 self.endWork();
                 console.error(err);
+                // does this return a rejected promise?
                 return err;
```
andrayantelo: @johntyree

docs/checkit.js
```diff
@@ -196,6 +201,8 @@ CheckIt.prototype.displayActiveCalendar = function() {
                         return checkit.displayCalendar(calendar);
                     }
                 })
+        // this gets run if getActive returns a rejected promise, but does it also 
+        // run if loadById returns a rejected promise?
```

docs/checkit.js
```diff
@@ -175,26 +175,40 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
+            return checkit.store.loadById(activeCalendarId)
+                // if loadById returns a rejected promise, TODO
+                .catch(function(err) {
+                    console.log("There is no current active calendar " + err);
+                    checkit.store.removeActive();
+                    checkit.showForm(checkit.$buildCalendarForm);
+                    return err;
+                });
         })
+                .then(function(activeCalendarState) {
```
johntyree: It looks like this is attached to `getActive()`, is that what you intended?

`getActive().then(....).then(....).catch(..);` ?

docs/checkit.js
```diff
@@ -685,30 +699,51 @@ CheckIt.prototype.deleteCalendar = function() {
...
+    }
+
+    var checkit = this;
+
+    return checkit.store.getActive()
+        // if rejected promise returned catch gets run
+        .catch(function(err) {
+            console.error("Couldn't get active calendar id: " + err);
+            // we don't return a resolved promise here, so after this, 
+            // we exit the function?
```
johntyree: Even if you *were* returning a rejected promise here, you would exit this function at this point.

docs/checkit.js
```diff
@@ -685,30 +699,51 @@ CheckIt.prototype.deleteCalendar = function() {
...
+        })
+        // if getActive promise resolved .then gets run
+        // with the resolved value from getActive
+        .then(function(currentCalendarId) {
```
johntyree: This is going to run whether or not `getActive()` succeeded because the `.catch()` above will "fix" it. By the time you get to this `.then()` it will think it "succeeded."

Further, since you don't even return a value from the `catch()` on 712, when this `.then()` gets called, `currentCalendarId` will be `undefined`.

docs/checkit.js
johntyree: So what you'd really like is for errors that can recover gracefully, but not if that means that we misrepresent things to the user. We wouldn't want to say we deleted the calendar, for example, and then have the user refresh the page and discover that we didn't. This could happen if e.g. their internet connection was unreliable and the call to remove failed.

Perhaps most importantly, we don't want to end up in a state where the database doesn't make sense anymore. As long as we avoid that, then it's not *that* big of a deal.

Do you have a way to [tell the user](https://notifyjs.com/) that there was an error of some kind?

johntyree (changes requested):

docs/checkit.js
```diff
@@ -711,26 +709,27 @@ CheckIt.prototype.deleteCalendar = function() {
...
         .catch(function(err) {
             console.error("Couldn't get active calendar id: " + err);
-            // we don't return a resolved promise here, so after this, 
-            // we exit the function?
+            // Notify user that there is no calendar on display for them
+            // to delete? Should I bother notifying them of that?
+            $.notify("No Calendar On Display");
+            // exit function, does returning err cause us to exit the function?
+            // or do I need to return a rejected promise?
+            return err;
```

docs/checkit.js
```diff
@@ -739,6 +738,7 @@ CheckIt.prototype.deleteCalendar = function() {
                     return checkit.store.removeActive()
                         .catch(function(err) {
                             console.error("Unable to remove active status from calendar");
+                            // Should I notify the user of something here?
                             return Promise.reject(err);
```

docs/checkit.js
```diff
@@ -186,23 +186,22 @@ CheckIt.prototype.displayActiveCalendar = function() {
...
+                var calendar = new Calendar(state);
+                return checkit.displayCalendar(calendar);
+            }
+        })
         // this gets run if getActive returns a rejected promise, but does it also 
-        // run if loadById returns a rejected promise?
+        // run if loadById returns a rejected promise??
         .catch(function(err) {
             console.log("Could not display active calendar" + err);
             checkit.showForm(checkit.$buildCalendarForm);
```
andrayantelo: So what do you think about the chaining in this function? @johntyree

docs/checkit.js
johntyree: Returning err will cause the next `.then()` in the chain to run with `err` as its argument.

docs/store.js
```diff
@@ -258,9 +258,12 @@ var FirebaseCalendarStorage = function(params) {
...
         
         self.startWork();
-        return new Promise(function(resolve, reject) {
+        return new Promise(/*executor*/function(resolve, reject) {
+            // TODO analyze this, .once returns a promise? we don't
+            // have to actually write 'return'?
             self.database.ref('users/' + userId + '/currentActiveCalendar')
-            .once('value')  
+            .once('value')
+            // currentActiveCalendar is the value returned by the .once promise?
```
johntyree: It is the value that that promise resolves to, yes.

docs/store.js
johntyree: Not as written. You'd need to `throw` or return `Promise.reject(err)`.

johntyree (approved):

PR #195, opened 2017-05-23
[WIP] Ability to edit title and/or start date and/or end date.

### LINKED ISSUE #30: Need ability to edit title, number of years, and startdate
Need to be able to edit these attributes after a calendar has been built and saved, but still preserve the checkmarks. 

Maybe just the title to start
Maybe use contenteditable?

"Edit Calendar" option in Options dropdown. 
When clicked, hide buildCalendar accordion, and have an edit calendar accordion?

andrayantelo:
Resolves #30 

This PR is about implementing the ability to edit a calendar's title, start date, and/or end date after building one. NOT about the CalendarAnalyzer class and its methods.

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -822,18 +822,22 @@ CheckIt.prototype.validateDates = function (startDateString, endDateString, $for
         difference = endDate.diff(startDate.format("YYYY-MM-DD"), 'years', true);
     
     if (startDate.isBefore(endDate)) {
+        console.log("startDate is before EndDate");
         // startDate is before endDate so errors can be removed.
         this.removeFieldError($formGroup, $sr);
         this.removeHelpBlock($helpBlock);
         // If there are more than 5 years between the dates return false for invalid
         if (difference < 5) {
+            console.log("start and end date are no more than five years apart");
             this.removeHelpBlock($fiveYears);
             return true;
         } else {
+            console.log("start date and end date are further than five years apart");
             this.addFieldError($formGroup, $sr);
             this.addHelpBlock($fiveYears);
         }
     } else {
+        console.log("startDate is after endDate");
         // Mark the endDate input field red because endDate is after startDate
         this.addFieldError($formGroup, $sr);
         this.addHelpBlock($helpBlock);
@@ -857,6 +861,7 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
     
     // If user hasn't written anything we fail immediately
     if (!inputVal) {
+        console.log("no input provided " + $srElement);
         this.addFieldError($inputFormGroup, $srElement);
         // if the input field has a span element with a glyphicon
         // reveal the error glyphicon
@@ -865,6 +870,7 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
         }
         return false;
     } else {
+        console.log("input valid " + $srElement);
         this.removeFieldError($inputFormGroup, $srElement);
         // Remove glyphicon if it exists
         if ($glyphicon.length) {
@@ -884,6 +890,8 @@ CheckIt.prototype.validateForm = function (startDateString, endDateString) {
         validateEndDate = this.validateInput(this.$fullForm, this.$endDateFormGroup, 'endDate'),
     
         isValid = validateTitle && validateStartDate && validateEndDate;
+    
+    console.log("boolean value for validateForm: " + isValid && this.validateDates(startDateString, endDateString, this.$endDateFormGroup));
     
     // Make sure input is OK before parsing and validating dates
     return (isValid && this.validateDates(startDateString, endDateString, this.$endDateFormGroup));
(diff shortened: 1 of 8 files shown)


docs/checkit.js
```diff
@@ -130,6 +130,102 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    "use strict";
+    return this.getTotalCalendarDays() - this.getNumberOfChecked();
+};
+
+CalendarAnalyzer.prototype.getCheckedDaysStreak = function () {
+    // Returns the longest streak of checked days
+    "use strict";
+    
+    var currentStreak = 0,
+        checkedDaysArray = [],
```
johntyree, line 166: This array never gets used. You might as well move the initializer up from line 175.

docs/checkit.js
```diff
@@ -130,6 +130,102 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+        i,
+        currentDay,
+        previousDay,
+        dayDiff;
+    
+    // place checkedDays object's keys into array and sort the array.
+    checkedDaysArray = Object.keys(this.calState.checkedDays);
+    checkedDaysArray.sort();
+
+    lastDate = checkedDaysArray[0];
```
johntyree, line 178: Unused? Also the name seems wrong. More like `firstDate`.

docs/checkit.js
```diff
@@ -130,6 +130,102 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    
+    for (i = 1; i < checkedDaysArray.length; i++) {
+        // if the next element is the day after lastDate, add 1 to currentStreak
```
johntyree, line 181: You don't use lastDate anywhere in here.

docs/checkit.js
```diff
@@ -130,6 +130,102 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    return bestStreak;
+};
+
+CalendarAnalyzer.prototype.getUncheckedDaysStreak = function () {
+    // Returns the longest streak of unchecked days
+    "use strict";
+    // TODO
+    // Go through all the dates between startDate and endDate, if they
+    // are not in checkedDays add them to an uncheckedDays array
+    // then find the bestStreak in that uncheckedDays array.
```
johntyree, line 206: Reinterpreting a new problem as a problem you already know to solve is a good strategy. In this particular case I believe there's an easier way. The largest gap between any two checked days (minus one) is all you need I think.

docs/checkit.js
```diff
@@ -130,6 +130,102 @@ Calendar.prototype.generateMonthObjects = function (startDate, endDate) {
...
+    // Returns the longest streak of unchecked days
+    "use strict";
+    // TODO
+    // Go through all the dates between startDate and endDate, if they
+    // are not in checkedDays add them to an uncheckedDays array
+    // then find the bestStreak in that uncheckedDays array. 
+
+};
+
+CalendarAnalyzer.prototype.getTotalCalendarWeeks = function () {
```
johntyree, line 210: This function is exactly like `getTotalCalendaDays`. Make a `getCalendarDuration` function that takes a unit (`'days'`, `'weeks'`, ...) as an argument and call that. These two can be one-line functions.

docs/checkit.js
```diff
@@ -800,7 +900,15 @@ CheckIt.prototype.createCalendar = function () {
...
-        
+    
+    // TODO
+    // if someone clicked on save, to edit their title and/or start date and/or end date
+    // then the same calendar state needs to be used as the currently displayed calendar
+    // except with the edited components changed.
+    // So validateform checks that each entry is filled in and filled in correctly,
+    // but this is only needed when creating a brand new calendar
+    // If we are already looking at a calendar then we need to isolate just the components
+    // that were edited.
```
johntyree: Or just populate the form with the calendar-to-be-edited's data and treat "saving" the form the same way, regardless of whether or not the calendar already exists?

johntyree (changes requested):

docs/checkit.js
andrayantelo: You're looking at the wrong PR for this particular problem.

andrayantelo:
@johntyree The comments you left in this PR, I'm applying to PR #186 .  Since they have to do with the CalendarAnalyzer and its methods.

docs/checkit.js
andrayantelo: What? lol
I mean, I'd need to have the same checkedDays... unless they picked a start date that comes after a bunch of checked days.. then I'd have to not have those anymore I guess.

docs/checkit.js
```diff
@@ -906,10 +906,17 @@ CheckIt.prototype.createCalendar = function () {
...
+    
     // If we are already looking at a calendar then we need to isolate just the components
     // that were edited. 
     
+    // Maybe I can prepopulate the buildCalendarForm with the currently displayed
+    // calendar's title, startDate, and endDate, so that if one of them gets changed, then
+    // we have all three pieces of data to do this step.
+    // Although, we'd need to get the currently displayed calendar's checkedDays data too.
+    // We could do all this in a new function called editCalendar?
+
```
andrayantelo: What do you think @johntyree?

PR #196, opened 2017-07-27
Build/edit calendar

### LINKED ISSUE #197: Tests for new functions
- [x] editOrCreate()
- [x] editCalendar()
- [x] preFillForm()

andrayantelo:
Build and edit use the same form. Resolves #197 and #30 and #199 and #201

### DIFF
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -843,16 +843,13 @@ CheckIt.prototype.editCalendar = function (title, start, end, activeCalId) {
     
     // I want to return a non-pending promise
     
-    loadP = checkit.store.loadById(activeCalId)
+    return checkit.store.loadById(activeCalId)
         .then(function (calState) {
             calState.endDateString = endDate.format("YYYYMMDD");
             calState.startDateString = startDate.format("YYYYMMDD");
             calState.title = title;
             checkit.createCalendar(calState);
-        
-            // TODO is this returning a promise?
         });
-    return Promise.all([loadP]);
 };
 
 
@@ -861,7 +858,6 @@ CheckIt.prototype.editOrCreate = function (params) {
     // or create a new one
     
     "use strict";
-    console.log("Running editOrCreate function");
     
     var checkit = this,
         state,
@@ -869,30 +865,31 @@ CheckIt.prototype.editOrCreate = function (params) {
         end = params.end,
         title = params.title;
     
-    console.log("With " + title + " " + start + " " + end);
     if (checkit.validateForm(start, end)) {
         
         // Check if there is a current active calendar
         // Don't have to return promise here? TODO
-        checkit.store.getActive()
+        return checkit.store.getActive()
             .then(function (activeCalId) {
                 if (activeCalId) { // I don't think I need this condition
                     // we are editing an existing calendar
-                    console.log("Editing calendar");
+                    
                     return checkit.editCalendar(title, start, end, activeCalId);
                 }
             }, function (err) { // Runs if getActive fails
                 // we are creating a new calendar
-                console.log("Creating calendar");
+                
                 state = emptyCalendarState({startDate: start, endDate: end, calendarTitle: title});
-                console.log("Creating calendar with " + JSON.stringify(state));
+                
                 return checkit.createCalendar(state);
             })
-            .catch(function (err) { // runs if either getActive fails or
+            .catch(function (err) { // runs if either getActive fails
                                     // or .then fails 
-                console.log("catch all error " + err);
+                
                 return Promise.reject(err);
             });
+    } else {
+        return Promise.reject("Invalid input.");
     }
 };
 
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -792,25 +792,7 @@ CheckIt.prototype.validateForm = function (startDateString, endDateString) {
    
 };
 
-CheckIt.prototype.editCalendar = function (title, start, end, activeCalId) {
-    // Edits the calendar currently on display (current active calendar)
-    // and displays the edited version
-    "use strict";
-    console.log("Running editCalendar function");
-    
-    var checkit = this,
-        startDate = moment(start, "YYYY-MM-DD"),
-        endDate = moment(end, "YYYY-MM-DD");
-    
-    return checkit.store.loadById(activeCalId)
-        .then(function (calState) {
-            calState.endDateString = endDate.format("YYYYMMDD");
-            calState.startDateString = startDate.format("YYYYMMDD");
-            calState.title = title;
-        });
-};
-
-CheckIt.prototype.createCalendar = function (title, start, end, calState) {
+CheckIt.prototype.createCalendar = function (calState) {
     // creates a calendar
     
     "use strict";
@@ -818,7 +800,6 @@ CheckIt.prototype.createCalendar = function (title, start, end, calState) {
         initP,
         buildP;
     console.log("Running createCalendar function");
-    console.log("Creating calendar with " + start + " " + end + " " + title);
     //if (this.validateForm(start, end)) {
         
     //clear the previously displayed calendar
@@ -843,6 +824,26 @@ CheckIt.prototype.createCalendar = function (title, start, end, calState) {
     //}
 };
 
+CheckIt.prototype.editCalendar = function (title, start, end, activeCalId) {
+    // Edits the calendar currently on display (current active calendar)
+    // and displays the edited version
+    "use strict";
+    console.log("Running editCalendar function");
+    
+    var checkit = this,
+        startDate = moment(start, "YYYY-MM-DD"),
+        endDate = moment(end, "YYYY-MM-DD");
+    
+    return checkit.store.loadById(activeCalId)
+        .then(function (calState) {
+            calState.endDateString = endDate.format("YYYYMMDD");
+            calState.startDateString = startDate.format("YYYYMMDD");
+            calState.title = title;
+            checkit.createCalendar(calState);
+        });
+};
+
+
 CheckIt.prototype.editOrCreate = function () {
     // Determine whether to edit an existing calendar
     // or create a new one
@@ -866,17 +867,14 @@ CheckIt.prototype.editOrCreate = function () {
         // Don't have to return promise here? TODO
         checkit.store.getActive()
             .then(function (activeCalId) {
-                console.log("Boolean(activeCalId) = " + Boolean(activeCalId));
                 if (activeCalId) { // I don't think I need this condition
                     // we are editing an existing calendar
                     return checkit.editCalendar(title, start, end, activeCalId);
                 }
             }, function (err) { // Runs if getActive fails
-                console.log("getActive error " + err);
-            
                 // we are creating a new calendar
                 state = emptyCalendarState({startDate: start, endDate: end, calendarTitle: title});
-                return checkit.createCalendar(title, start, end, state);
+                return checkit.createCalendar(state);
             })
             .catch(function (err) { // runs if either getActive fails or
                                     // or .then fails 
@@ -964,6 +962,16 @@ CheckIt.prototype.addCalendarToDropdown = function (uniqueId, title, $dropdown) 
     //add the calendar with unique Id, uniqueId, and title, title, to
     //the saved calendars dropdown on the navbar.
     "use strict";
+    
+    // first check if the title isn't already in the dropdown
+    if ($('#' + uniqueId).length) {
+        // title is already in dropdown
+        // but the title may have changed
+        // so just update the title
+        $('#' + uniqueId).find("a").html(title);
+        return;
+    }
+    
     $dropdown.append('<li id="' + uniqueId
             + '"><a href=#>' + title + '</a></li>');
 };
--- a/docs/checkit.js
+++ b/docs/checkit.js
@@ -804,8 +804,8 @@ CheckIt.prototype.editCalendar = function (title, start, end, activeCalId) {
         checkit = this;
     stateP = checkit.store.loadById(activeCalId);
     
-    Promise.all([stateP]).then(function (val) {
-        
+    Promise.all([stateP]).then(function (calState) {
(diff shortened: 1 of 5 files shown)


docs/checkit.js
```diff
@@ -342,7 +342,7 @@ CheckIt.prototype.displayActiveCalendar = function () {
             }
         })
         // this gets run if getActive returns a rejected promise, but does it also 
-        // run if loadById returns a rejected promise??
+        // run if loadById returns a rejected promise?
```
andrayantelo: Does it? @johntyree

docs/checkit.js
```diff
@@ -823,6 +823,12 @@ CheckIt.prototype.createCalendar = function () {
                 checkit.editCalendar();
             }
         });
+    // Don't need to catch here, because if it fails
+    // we just go on with the createCalendar function
+
```
andrayantelo: Right? @johntyree

docs/checkit.js
```diff
@@ -810,14 +810,15 @@ CheckIt.prototype.createCalendar = function () {
...
+        state,
+        calendar,
+        initP,
+        buildP,
         checkit = this;
     
     // Check if there is a current active calendar
-    return checkit.store.getActive()
+    // Don't have to return promise here?
+    checkit.store.getActive()
```

docs/checkit.js
```diff
@@ -821,16 +818,12 @@ CheckIt.prototype.createCalendar = function () {
     checkit.store.getActive()
         .then(function (activeCalId) {
             if (activeCalId) {
-                checkit.editCalendar();
+                checkit.editCalendar(title, start, end);
             }
+        })
+        .catch(function (err) {
+            console.log("no active calendar " + err);
         });
```
andrayantelo: getActive returns an error that I need to catch. @johntyree

docs/checkit.js
```diff
@@ -854,20 +851,26 @@ CheckIt.prototype.editOrCreate = function () {
...
+        }, function (err) { // Runs if getActive fails
+            console.log("getActive error " + err);
+            state = emptyCalendarState({startDate: start, endDate: end, calendarTitle: title});
+            return checkit.createCalendar(title, start, end, state);
         })
-        .catch(function (err) {
-            console.log("could not create/edit calendar " + err);
+        .catch(function (err) { // runs if either getActive fails or
+                                // or .then fails 
+            console.log("catch all error " + err);
```
andrayantelo: Does .then ever fail? @johntyree

docs/checkit.js
johntyree: I don't know. I could spend time figuring it out, but how would you?

docs/checkit.js
```diff
@@ -341,8 +344,8 @@ CheckIt.prototype.displayActiveCalendar = function () {
                 return checkit.displayCalendar(calendar);
             }
         })
-        // this gets run if getActive returns a rejected promise, but does it also 
-        // run if loadById returns a rejected promise??
+        // this gets run if getActive returns a rejected promise, but does it // also 
+        // run if loadById returns a rejected promise?
```
johntyree, line 348: I would expect it to, but you can test it yourself easily.

docs/checkit.js
```diff
@@ -732,9 +739,11 @@ CheckIt.prototype.validateDates = function (startDateString, endDateString, $for
         } else {
             this.addFieldError($formGroup, $sr);
             this.addHelpBlock($fiveYears);
+            // Shouldn't I be returning false here?
```
johntyree, line 742: Yes. I'd probably do something like this.

```js

    ...

     var startDate = moment(startDateString, "YYYY-MM-DD"),
         endDate = moment(endDateString, "YYYY-MM-DD"),
         $sr = $formGroup.find('.sr-only'),
         $helpBlock = $formGroup.find('#helpBlock'),
         $fiveYears = $formGroup.find('#fiveYears'),
         difference = endDate.diff(startDate.format("YYYY-MM-DD"), 'years', true),
         isValid = true;

    this.removeFieldError($formGroup, $sr);
    this.removeHelpBlock($helpBlock);
   
    if (endDate.isBefore(startDate)) {
        isValid = false;
        // Mark the endDate input field red
        console.log("endDate is before startDate");
        this.addFieldError($formGroup, $sr);
        this.addHelpBlock($helpBlock);
    }
    
    // Don't allow more than a 5 year span
    if (difference > 5) {
        isValid = false;
        this.addHelpBlock($fiveYears);
    }

    return isValid;
};
```

docs/checkit.js
```diff
@@ -777,55 +786,118 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
...
-    var title = this.$calendarTitle.val(),
-        start = this.$startDate.val(),
-        end = this.$endDate.val(),
-        state,
-        calendar,
+    var calendar,
         initP,
         buildP;
-        
+    //if (this.validateForm(start, end)) {
```
johntyree, line 811: Why is this commented out?

johntyree (changes requested):
Not done yet, but here's some.

docs/checkit.js
andrayantelo: OKay, but I think I will do this in a separate branch

docs/checkit.js
```diff
@@ -777,55 +786,118 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
...
+    // Edits the calendar currently on display (current active calendar)
+    // and displays the edited version
+    "use strict";
+    
+    var checkit = this,
+        startDate = moment(start, "YYYY-MM-DD"),
+        endDate = moment(end, "YYYY-MM-DD"),
+        loadP;
+    
+    // I want to return a non-pending promise
```
johntyree, line 845: What do you mean here? Promises are pending until they resolve. You don't decide whether or not it is pending. It will be pending until the work is done.

docs/checkit.js
andrayantelo: I have no idea.

andrayantelo: Oh wait, I know. Because when you click on "create calendar" on the website, the function editOrCreate gets run, and in there the form is validated and based on which button was pressed either editCalendar gets run or createCalendar.

docs/checkit.js
```diff
@@ -777,55 +786,118 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
...
+        .then(function (calState) {
+            calState.endDateString = endDate.format("YYYYMMDD");
+            calState.startDateString = startDate.format("YYYYMMDD");
+            calState.title = title;
+            checkit.createCalendar(calState);
+        });
+};
+
+
+CheckIt.prototype.editOrCreate = function (params) {
```
johntyree, line 857: I think this function does too much work and is hard to understand.

Seems like what you want is a boolean function that tells you whether or not there is an active calendar (which would always imply that it's displayed) and then based on that create one or not. Then edit it if needed.

docs/checkit.js
```diff
@@ -777,55 +786,118 @@ CheckIt.prototype.validateInput = function ($form, $inputFormGroup, inputId) {
...
+                    if (activeCalId) { // I don't think I need this condition
+                        // we are editing an existing calendar
+
+                        return checkit.editCalendar(title, start, end, activeCalId);
+                    } else {
+                        console.log("No calendar currently on display");
+                        return Promise.reject("No calendar currently on display");
+                    }
+                })
+                .catch(function (err) { // runs if either getActive fails
```
johntyree, line 889: I don't think you need this catch if all it's going to do is return another rejected promise. It's already a rejected promise if you're inside the catch.

docs/checkit.js
```diff
@@ -907,6 +981,16 @@ CheckIt.prototype.addCalendarToDropdown = function (uniqueId, title, $dropdown)
     //add the calendar with unique Id, uniqueId, and title, title, to
     //the saved calendars dropdown on the navbar.
     "use strict";
+    
+    // first check if the title isn't already in the dropdown
+    if ($('#' + uniqueId).length) {
```
johntyree, line 986: Why length?

docs/checkit.js
```diff
@@ -907,6 +981,16 @@ CheckIt.prototype.addCalendarToDropdown = function (uniqueId, title, $dropdown)
...
+        // title is already in dropdown
+        // but the title may have changed
+        // so just update the title
+        $('#' + uniqueId).find("a").html(title);
```
johntyree, line 990: Probably want to sanitize `title` and not just stuff it in as HTML. Does `innerText` or something do it? I don't remember. Anyway, you never want to let the browser evaluate unsanitized user input, be it JS, HTML, CSS, whatever.

docs/checkit.js
andrayantelo: It does.

docs/checkit.js
```diff
@@ -1166,4 +1286,4 @@ CheckIt.prototype.opts = {
     shadow: false, // Whether to render a shadow
     hwaccel: false, // Whether to use hardware acceleration
     position: 'absolute' // Element positioning
-};
+};
```
johntyree, line 1289: Missing a newline

johntyree (changes requested):

docs/checkit.js
andrayantelo: I don't know why I wrote that.

andrayantelo: Why return a promise there at all?

docs/checkit.js
andrayantelo: Yeah but it works. So. Who cares? I thought that's how things were done. 😆

docs/checkit.js
andrayantelo: 👍

docs/checkit.js
andrayantelo: I'm not sure.

docs/checkit.js
andrayantelo: What do you mean sanitize?

docs/checkit.js
andrayantelo: what do you mean?

docs/checkit.js
andrayantelo: The answer is yes.

docs/checkit.js
andrayantelo: I guess I can just take away .length and have 
`if ($('#' + uniqueId))`

right?

andrayantelo:
Abandoning this branch because of the way it is written.

docs/checkit.js
johntyree: Because the caller is probably expecting this function to return a Promise, I would imagine? If so, then if you don't return a Promise it will break when it tries to resolve it.

docs/checkit.js
johntyree: You can definitely just leave it if it works. It will just be harder to understand the next time you come look at it to fix a bug or update functionality.

docs/checkit.js
johntyree: I'm not sure. What does `$('#foo')` return if there is no element with id `foo` ? Does JS treat whatever that is as `true` when used in an `if` statement?

You can test it in the console pretty easily. You can also explain exactly why it's like this in more detail in the comment.

johntyree: Looks like `length` is indeed needed. Otherwise JS looks at the jquery object and objects are `true`.

```javascript
> $('#foo')
jQuery.fn.init {context: document, selector: "#foo"}
>$('#foo').length
0
> Boolean($('#foo'))
true
> Boolean($('#foo').length)
false
```

docs/checkit.js
johntyree: In general, sanitization is the process of taking untrusted data (like from a user) and attempting to make it safe to use as an input to something else (like displaying on a web page. It's hard to do correctly most of the time. It's the reason you write e.g. SQL queries using `?` in place of variables and let the sql engine itself do the substitution. If you were to just build the whole query string yourself, but I'm allowed to add arbitrary text to the middle of it because I control the value of a variable you're using, I can write something clever that will screw everything up. The SQL `execute()` function will make sure that what I wrote is properly escaped and safe before  it runs.

In this case, you're taking the title of the calendar, which the user has total control over, and stuffing it into an HTML node in the DOM, which means the browser is going to evaluate it. I could make my title be something crazy like a `<script>` tag that contains javascript that uploads random garbage data until your app runs out of space or you get a giant bill from Google for resource usage, etc.

I think what you want here is [createTextNode](https://developer.mozilla.org/en-US/docs/Web/API/Document/createTextNode).

Maybe something like this (untested)?

```js
var $elem = $('#' + uniqueId).find('a');
$elem.empty()
$elem.createTextNode(title);
```

docs/checkit.js
johntyree: Because of the way that newline characters are used to indicate the end of the line it is expected that every file ends with a newline character.

Github showed a little warning that this file didn't have one (:no_entry_sign:).

johntyree:
I see that it's abandoned, but I left comments anyway.

docs/checkit.js
andrayantelo: I opened up an issue for this just now.

docs/checkit.js
andrayantelo: Like I have to go down to the bottom of the page and press enter?

johntyree: Basically, yes.

A file is just a blob of bytes. If the file is supposed to represent text (as opposed to say... an image) then the convention is that the final line of text will end with a newline character so that if you were to print it out, the cursor would move back to column 0 on the screen, ready to start printing the next file.

For example here's a text file with all of the whitespace characters made visible:
```
This is a text file ending in a newline. That's how you know it's complete.\n
```

Here is one without. It makes it difficult to know if the file is actually complete or not.

```
This is a text file that doesn't end in a newline. Maybe part of it was cut off or maybe this is just how it
```

Because of this, most code linters (including github's) will complain about files that don't end in a newline.

